PR #11, opened 2021-04-22
Feature/add login endpoint

AWitcherONS:
### What

- Endpoint for Tokens 
- logic for 400 error response for invalid email or password
- Unit for above
- feature test for above
- swagger for structures from Tokens Endpoint
this pull does not include integration with Cognito that would handle 201 response

### How to review

make test and make test-component

### Who can review

anyone


api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+type IndividualError struct {
+	SpecificError string `json:"error"`
+	Message       string `json:"message"`
+	Source        Source `json:"source"`
+}
+
+type Source struct {
+	Field string `json:"field"`
+	Param string `json:"param"`
+}
```
daiLlew: Public types should have GoDoc.

Question: If this is going to be the new API standard should this live in a library? I am guessing it was decided that is beyond the scope of this task and will be something we do later but I thought I'd ask now anyway.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+		param := ""
+
+		body, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			errorMessage := "api endpoint POST login returned an error reading the request body"
+			handleUnexpectedError(ctx, w, err, errorMessage, field, param)
+			return
+		}
+
+		defer req.Body.Close()
```
daiLlew: I think this should go before line 39. I'm not 100% sure but my _coding spidey-senses_ make me think it's safer to register the defer close before using it. I am not sure if `ioutil.ReadAll` will close the body so best to err on the side of caution.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+			return
+		}
+
+		validPasswordRequest := passwordValidation(authParams)
+		validEmailRequest := emailValidation(authParams)
+		if err != nil {
+			errorMessage := "api endpoint POST login returned an error validating the email"
+			handleUnexpectedError(ctx, w, err, errorMessage, field, param)
+			return
+		}
```
daiLlew: I'm not sure this correct? should this be checking `emailValidation`? As it is `err` will have the value assigned on line `49`.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+
+		invalidPasswordError := errors.New("Invalid password")
+		invalidPasswordMessage := "Unable to validate the password in the request"
+		invalidEmailError := errors.New("Invalid email")
+		invalidErrorMessage := "Unable to validate the email in the request"
```
daiLlew: Consider making these to package vars - they will always be the same values so it makes sense to create once and reuse. Also I think the Go standards around errors says that should be lower cause (not 100% sure though) I will see if I can find a link.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+
+		invalidPasswordErrorBody := individualErrorBuilder(invalidPasswordError, invalidPasswordMessage, field, param)
+		invalidEmailErrorBody := individualErrorBuilder(invalidEmailError, invalidErrorMessage, field, param)
+
+		var errorList []IndividualError
+
+		if !(validPasswordRequest) && !(validEmailRequest) {
```
daiLlew: The parenthesises aren't required.  Consider simplifying this expression, I think you could do something like:

```
if validPasswordRequest && validEmailRequest {
    // Do success request steps.
    // return success response.
}

// otherwise check the fields and create the appropriate errors.
// return appropriate error response.
}

File: api/tokens.go [hunk 6]
```diff
@@ -0,0 +1,179 @@
...
+
+			errorList = nil
+			errorList = append(errorList, invalidPasswordErrorBody)
+			errorList = append(errorList, invalidEmailErrorBody)
```
daiLlew: I think this will cause a nil pointer error when append is called as `errorList` is nil. I think you just need to remove `errorList = nil`.

Also I think you can do this:
````
errorList = append(errorList, invalidPasswordErrorBody, invalidEmailErrorBody)
````

File: api/tokens.go [hunk 7]
```diff
@@ -0,0 +1,179 @@
...
+
+			errorResponseBody := errorResponseBodyBuilder(errorList)
+			writeErrorResponse(ctx, w, 400, errorResponseBody)
+			return
+		}
+
+		if !validPasswordRequest {
+
+			errorList = nil
```
daiLlew: Same here.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+
+			writeErrorResponse(ctx, w, 400, errorResponseBody)
+			return
+		}
+	}
+}
+
+func passwordValidation(requestBody map[string]string) (isPasswordValid bool) {
+
+	isPasswordValid = false
```
daiLlew: I don't think this is required. When using a named return the variable is initialised with it's _zero value_ which for a boolean is false.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if len(requestBody["password"]) != 0 {
+		isPasswordValid = true
+	}
```
daiLlew: You could probably simplify this to:
```
return len(requestBody["password"]) > 0
``` 
😄

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func TokensHandler() http.HandlerFunc {
+
+	return func(w http.ResponseWriter, req *http.Request) {
+
+		ctx := req.Context()
+
+		field := ""
+		param := ""
```
daiLlew, line 33: This is absolutely fine but I think the Go convention is:

`var field string`

When you declare a var this way it will have the _zero value_ for that type which for String will be "".  As I said nothing technically wrong with this.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+			errorList = append(errorList, invalidPasswordErrorBody)
+			errorResponseBody := errorResponseBodyBuilder(errorList)
+
+			writeErrorResponse(ctx, w, 400, errorResponseBody)
+			return
+		}
+
+		if !validEmailRequest {
+
+			errorList = nil
```
daiLlew: Same here.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+		body, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			errorMessage := "api endpoint POST login returned an error reading the request body"
+			handleUnexpectedError(ctx, w, err, errorMessage, field, param)
+			return
+		}
+
+		defer req.Body.Close()
+
+		authParams := make(map[string]string)
```
daiLlew: I think it would be worthwhile creating a type to hold the body data. A map works but it's very open and you could in theory send any JSON data to the endpoint and it would end up being Unmarshalled here.

The other benefit of having a type is that you _could_ (if you wanted to) add validation functions as methods on that type - the data belongs to the model so the model knowing what it is valid feels like reasonable functionality. This is just a suggestion and I'm not specifically requesting this change - I'm just offering food for thought.

However, I do think creating a struct for the request body is necessary.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func errorResponseBodyBuilder(listOfErrors []IndividualError) (errorResponseBody ErrorStructure) {
+
+	errorResponseBody = ErrorStructure{
+		Errors: listOfErrors,
+	}
+
+	return errorResponseBody
+}
```
daiLlew: Consider moving the Error types and these functions into a separate package. It feels like it is its own separate thing, this will also make it easier if/when we extract this into it's own library,

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+
+func writeErrorResponse(ctx context.Context, w http.ResponseWriter, status int, errorResponseBody interface{}) {
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+
+	jsonResponse, _ := json.Marshal(errorResponseBody)
```
daiLlew: You need to assign and handle the error returned by this func.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+	_, err := w.Write(jsonResponse)
+	if err != nil {
+
+		log.Event(ctx, "writing response failed", log.Error(err), log.ERROR)
+		http.Error(w, "Failed to write http response", http.StatusInternalServerError)
+		return
+	}
+
+	return
```
daiLlew: The return is not required here - it will happen anyway.

api/tokens.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func handleUnexpectedError(ctx context.Context, w http.ResponseWriter, err error, message, sourceField, sourceParam string) {
+
+	var errorList []IndividualError
+
+	errorList = nil
```
daiLlew: Remove this line - see previous comment.

emailvalidation/emailvalidation.go
```diff
@@ -0,0 +1,25 @@
+package emailvalidation
```
daiLlew: This feels very very specific. Maybe we could broaden it a bit and make this a `validation` package? That way it will serve a specific purpose (validation) but won't be so specific it is only used in 1 case.

emailvalidation/emailvalidation.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"net"
+	"regexp"
+	"strings"
+)
+
+var emailRegex = regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
```
daiLlew: Consider adding some GoDoc here to outline what the pattern is looking for - it's helpful to others in future and also reminds you what this is doing 6months down the line when you've totally forgetting 😂

emailvalidation/emailvalidation.go
```diff
@@ -0,0 +1,25 @@
...
+
+//IsEmailValid is a function to validate email addresses
+func IsEmailValid(e string) bool {
+	if len(e) < 3 && len(e) > 254 {
```
daiLlew: 3 and 254 are [magic numbers](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad). Create vars with descriptive names to improve readability.

emailvalidation/emailvalidation.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"net"
+	"regexp"
+	"strings"
+)
+
+var emailRegex = regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
+
+//IsEmailValid is a function to validate email addresses
```
daiLlew: Probably worth adding what is considered a valid email to the GoDoc - helpful to anyone using this func.

daiLlew (changes requested):
From my initial review the business logic of what you doing looks sensible. I've added a couple of comments for changes and suggestions, I'm happy to have a chat to discuss if you have any questions.

api/tokens.go
HeJones: Thank you for pointing this out. 

Yes I think that it will live in a library but for now it was just decided to align the responses with the expected behaviour of the library.

api/tokens.go
HeJones: Well spotted, the emailValidation used to return an error but with the refactoring to create the emailValidation package it no longer does and removing this check was missed.

api/tokens.go
HeJones: Thank you, I wasn't too sure if these should be package vars.

api/tokens.go
HeJones: This is a good idea, after the success case would we still need to check if both are false to create the error in this scenario?

api/tokens.go
HeJones: Thank you for spotting this and the suggestion

api/tokens.go
HeJones: Thank you, I wasn't too sure about how to initialise the list

api/tokens.go
HeJones: Thank you for the suggestion

api/tokens.go
HeJones: Thank you for spotting this one

api/tokens.go
daiLlew: My general philosophy is - if something is dynamically generated then create when its called. If it's always the same regardless of input it feels like a "constant" or predefined package variable.

api/tokens.go
daiLlew: In the above both have to be true for it to be successful. If that condition is not met then 1 or both must be false so I think you would check 1 at a time adding an errors to the error response. You would `return` only after checking both cases:
```
// if passwordInvalid {
    // handle invalid password
}

if emailInvalid {
    // handle invalid email
}

// return
```
Does that make sense?

api/tokens.go
daiLlew: Actually I was wrong. According to the [documentation](https://golang.org/pkg/builtin/#append) for the built in `append()` func if the slice doesn't have sufficient capacity to accommodate the new entry a new array is allocated. So this actually wouldn't nil pointer **but** the nil assign isn't required so we should still remove it.

Here is a [helpful resource on how to initialise an array](https://gobyexample.com/arrays). You can also use the build in make func to create an array with the specified capacity:
```
myList := make([]int, 3) // new empty array with capacity of 3 i.e. [0 0 0]
```
[This post](https://medium.com/@habibridho/golang-nil-vs-empty-slice-87fd51c0a4d) explains nil vs empty arrays better than I can. Hopefully this is helpful.

api/tokens.go
HeJones: Updated so the defer is above the `ioutil.ReadAll`

api/tokens.go
HeJones: This has now been removed

api/tokens.go
HeJones: These variables have been moved, but they have been left lower case because at the moment they aren't used outside this package

api/tokens.go
HeJones: This has been updated now

api/tokens.go
HeJones: Thank you for the link, and comment. This has also been removed from the code

api/tokens.go
HeJones: Thank you for your help, the code has been updated

api/tokens.go
HeJones: This has been updated

api/tokens.go
HeJones: Thank you, this has been updated

api/tokens.go
HeJones: This has been removed

api/tokens.go
HeJones: A struct has been created and this is being used rather than a map

api/tokens.go
HeJones: These have been moved to a separate package now

api/tokens.go
HeJones: This has now been removed

api/tokens.go
HeJones: This error is now being handled in a similar way to the error that could occur when the error response couldn't be written

api/tokens.go
HeJones: This line has been removed

emailvalidation/emailvalidation.go
HeJones: These have been updated to variables

emailvalidation/emailvalidation.go
HeJones: I think there was the feeling that it might be worth having a specific package for email validation, because whilst specific it would be reused in more than one scenario?

emailvalidation/emailvalidation.go
HeJones: Documentation has been added to capture what has been considered a valid email

emailvalidation/emailvalidation.go
HeJones: This package has been updated to validation

api/tokens.go
```diff
@@ -0,0 +1,124 @@
...
+		if !validPasswordRequest {
+			errorList = append(errorList, invalidPasswordErrorBody)
+		}
+
+		if !validEmailRequest {
+			errorList = append(errorList, invalidEmailErrorBody)
+		}
+
+		errorResponseBody := apierrors.ErrorResponseBodyBuilder(errorList)
+		writeErrorResponse(ctx, w, 400, errorResponseBody)
```
mike-kelly-methods: Rather than inlining the `400` response, it's better to use `http.StatusBadRequest` (`net/http` package)

HeJones: This has been updated now, and also a 500 further down

emailvalidation/emailvalidation.go
HeJones: Documentation has been added to explain the regex. Thank you @mike-kelly-methods

mike-kelly-methods: 👍

api/tokens.go
daiLlew: I think this if err != nil block still needs to be removed

daiLlew (approved):

PR #12, opened 2021-04-26
Add describe user pool request handler for health check

mike-kelly-methods:
### What

As per title. Adds a health check endpoint that uses aws cognito client's `DescribeUserPool` api endpoint to determine the health of the cognito service and reports `CRITICAL` if an error is found. 

### How to review

Review code and run unit tests.

### Who can review

Anyone but me.


api/api.go
```diff
@@ -3,33 +3,23 @@ package api
 import (
 	"context"
 
-	"github.com/ONSdigital/dp-identity-api/config"
+	cognitoclient "github.com/ONSdigital/dp-identity-api/cognitoclient"
```
daiLlew: Not sure this alias is needed as the package name is the same as the alias. Also consider renaming the package as it _"stutters"_ i.e. `client.client` which is against the Go naming convention.  Go is very opinionated about naming things which can present an annoying burden to work around. `cognito` seems like a reasonable package name.

cognitoclient/cognito.go
```diff
@@ -0,0 +1,11 @@
+package cognitoclient
```
daiLlew: As mentioned consider renaming this package to avoid the stutter. I think `cognito` would be appropriate but whatever works is fine.

cognitoclient/mock/cognito.go
```diff
@@ -0,0 +1,24 @@
...
+// mocked functions
+
+func (m *MockCognitoIdentityProviderClient) DescribeUserPool(poolInputData *cognitoidentityprovider.DescribeUserPoolInput) (*cognitoidentityprovider.DescribeUserPoolOutput, error) {
+	exsitingPoolID := "us-west-2_aaaaaaaaa"
+    if *poolInputData.UserPoolId != exsitingPoolID {
+		return nil, errors.New("Failed to load user pool data")
+	} 
+	return nil, nil
+}
+
```
daiLlew: This is fine but it's as it's hardcoded it's not flexible (I thinking of future uses where we might need to test different scenarios). If you updated it to something like:
```
type MockCognitoIdentityProviderClient struct {
	cognitoidentityprovideriface.CognitoIdentityProviderAPI

	DescribeUserPoolFunc func(poolInputData *cognitoidentityprovider.DescribeUserPoolInput) (*cognitoidentityprovider.DescribeUserPoolOutput, error)
}

func (m *MockCognitoIdentityProviderClient) DescribeUserPool(poolInputData *cognitoidentityprovider.DescribeUserPoolInput) (*cognitoidentityprovider.DescribeUserPoolOutput, error) {
	return m.DescribeUserPoolFunc()
}
```
It can be customised to behave however is required. When you construct the mock you give it the function you want it to execute.
````
m := &MockCognitoIdentityProviderClient{
		DescribeUserPoolFunc := func(poolInputData *cognitoidentityprovider.DescribeUserPoolInput) (*cognitoidentityprovider.DescribeUserPoolOutput, error) {
			// set behaviour as reqiured here.
		}
	}
````
Not strictly necessary but food for thought.

File: config/config.go [hunk 4]
```diff
@@ -13,6 +13,7 @@ type Config struct {
 	HealthCheckInterval        time.Duration `envconfig:"HEALTHCHECK_INTERVAL"`
 	HealthCheckCriticalTimeout time.Duration `envconfig:"HEALTHCHECK_CRITICAL_TIMEOUT"`
 	AWSRegion                  string        `envconfig:"AWS_REGION"`
+	AWSCognitoUserPoolID       string        `envconfig:"AWS_COGNTIO_USER_POOL_ID"`
```
daiLlew: Is this potentially sensitive information? If so should we use the json omit tag so it's not displayed in logging.

service/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+const CognitoHealthy  = "Cognito Healthy"
+
+func CognitoHealthCheck(cognitoClient cognitoclient.Client, userPoolID *string) health.Checker {
+	
+	return func(ctx context.Context, state *health.CheckState) error {
+		_, err := cognitoClient.DescribeUserPool(&cognito.DescribeUserPoolInput{UserPoolId: userPoolID})
+	
+		if err != nil {
```
daiLlew, line 21: If its possible to get additional detail from the `err` it might be useful to log it. I'm thinking it might be useful if anyone has to investigate an unexpected issue - the more details available the better.

daiLlew (dismissed):

daiLlew (approved):

PR #16, opened 2021-05-06
add commands to get the client id and secret to the debug command

mwnicks:
### What

Added additional steps to the make debug command to set cognito credentials for local running

### How to review

Run `make debug`
See that the api starts
Check the print out for AWS_CLIENT_ID and AWS_CLIENT_SECRET to see they are being correctly set.

### Who can review

Anyone but me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ build:
 
 .PHONY: debug
 debug:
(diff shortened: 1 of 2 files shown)


sandypadmanabhan (approved):
we will need to update the secrets in dp-config as well, as the change `envconfig:"AWS_COGNITO_USER_POOL_ID" json:"-"`is in both develop and production

PR #19, opened 2021-05-11
Feature/user logout

mwnicks:
### What

Added sign out end point as per [Trello ticket ](https://trello.com/c/q4yrHxSX/294-create-logout-endpoint-in-dp-identity-api)

### How to review

Check the unit and feature tests are passing.

Generate an access token using the dp-auth-poc or AWS Cogntio API (message me if you need credentials to test with)
Add the generated access token to a DELETE request to localhost:25600/tokens/self
You should receive a 204 response

### Who can review

Anyone but me

### DIFF
--- a/api/tokens.go
+++ b/api/tokens.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ONSdigital/dp-identity-api/validation"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	"github.com/ONSdigital/log.go/log"
 )
@@ -82,7 +83,17 @@ func (api *API)tokensLogoutHandler(w http.ResponseWriter, req *http.Request) {
 	if authString == "" {
 		invalidTokenErrorBody := apierrors.IndividualErrorBuilder(invalidTokenError, invalidTokenMessage, field, param)
 		errorList = append(errorList, invalidTokenErrorBody)
-		log.Event(ctx, "no access token provided", log.ERROR)
+		log.Event(ctx, "no authorization header provided", log.ERROR)
+		errorResponseBody := apierrors.ErrorResponseBodyBuilder(errorList)
+		writeErrorResponse(ctx, w, http.StatusBadRequest, errorResponseBody)
+		return
+	}
+
+	authComponents := strings.Split(authString, " ")
+	if len(authComponents) != 2 {
+		log.Event(ctx, "malformed authorization header provided", log.ERROR)
+		invalidTokenErrorBody := apierrors.IndividualErrorBuilder(invalidTokenError, invalidTokenMessage, field, param)
+		errorList = append(errorList, invalidTokenErrorBody)
 		errorResponseBody := apierrors.ErrorResponseBodyBuilder(errorList)
 		writeErrorResponse(ctx, w, http.StatusBadRequest, errorResponseBody)
 		return
--- a/api/tokens.go
+++ b/api/tokens.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"github.com/ONSdigital/dp-identity-api/apierrors"
 	"github.com/ONSdigital/dp-identity-api/validation"
+	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
 	"io/ioutil"
 	"net/http"
 	"strings"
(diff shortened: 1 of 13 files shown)


api/tokens_test.go
```diff
@@ -134,3 +140,85 @@ func TestHandleUnexpectedError(t *testing.T) {
...
+
+	// mock call to: GlobalSignOut(input *cognitoidentityprovider.GlobalSignOutInput) (*cognitoidentityprovider.GlobalSignOutOutput, error)
+	m.GlobalSignOutFunc = func(signOutInput *cognitoidentityprovider.GlobalSignOutInput) (*cognitoidentityprovider.GlobalSignOutOutput, error) {
+		return &cognitoidentityprovider.GlobalSignOutOutput{}, nil
+	}
+
+	api := Setup(ctx, r, m, poolId, clientId, clientSecret)
+
+	Convey("Global Sign Out returns 204: successfully signed out user", t, func() {
+		r := httptest.NewRequest(requestType, signOutEndPoint, bytes.NewReader(nil))
```
mike-kelly-methods: Could use `http.MethodDelete` here and also can `nil` the `body` param here:

`r := httptest.NewRequest(requestType, signOutEndPoint, nil)`

api/tokens.go
```diff
@@ -71,6 +74,54 @@ func TokensHandler() http.HandlerFunc {
...
+		field := ""
+		param := ""
+
+		authString := req.Header.Get("Authorization")
+		if authString == "" {
+			invalidTokenErrorBody := apierrors.IndividualErrorBuilder(invalidTokenError, invalidTokenMessage, field, param)
+			errorList = append(errorList, invalidTokenErrorBody)
+			log.Event(ctx, "no authorization header provided", log.ERROR)
+			errorResponseBody := apierrors.ErrorResponseBodyBuilder(errorList)
+			writeErrorResponse(ctx, w, http.StatusBadRequest, errorResponseBody)
```
mike-kelly-methods, line 88: This is interesting; I'm appending all validation errors to the an `errorList []models.IndividualError` array and then returning them all rather than individually (as the ticket requires this) by doing:

`		// report validation errors in response`

`		if len(errorList) != 0 {
			apierrors.WriteErrorResponse(ctx, w, http.StatusBadRequest, apierrors.ErrorResponseBodyBuilder(errorList))
			return
		}`

api/tokens_test.go
```diff
@@ -134,3 +140,85 @@ func TestHandleUnexpectedError(t *testing.T) {
...
+
+		w := httptest.NewRecorder()
+
+		api.Router.ServeHTTP(w, r)
+
+		So(w.Code, ShouldEqual, http.StatusNoContent)
+	})
+
+	Convey("Global Sign Out returns 400: malformed header", t, func() {
+		r := httptest.NewRequest(requestType, signOutEndPoint, bytes.NewReader(nil))
```
mike-kelly-methods: ditto above.

api/tokens_test.go
```diff
@@ -134,3 +140,85 @@ func TestHandleUnexpectedError(t *testing.T) {
...
+
+		w := httptest.NewRecorder()
+
+		api.Router.ServeHTTP(w, r)
+
+		So(w.Code, ShouldEqual, http.StatusBadRequest)
+	})
+
+	Convey("Global Sign Out returns 400: no header value", t, func() {
+		r := httptest.NewRequest(requestType, signOutEndPoint, bytes.NewReader(nil))
```
mike-kelly-methods: ditto above.

api/tokens_test.go
```diff
@@ -134,3 +140,85 @@ func TestHandleUnexpectedError(t *testing.T) {
 		So(resp.Body.String(), ShouldResemble, errorResponseBodyExample)
 	})
 }
+
+func TestSignOutHandler(t *testing.T) {
+	var (
+		r                                                  = mux.NewRouter()
+		ctx                                                = context.Background()
+		requestType, poolId, clientId, clientSecret string = "DELETE", "us-west-11_bxushuds", "client-aaa-bbb", "secret-ccc-ddd"
```
mike-kelly-methods: If you use the `http.MethodDelete` for `DELETE` method, can get rid of `requestType`.

api/tokens_test.go
```diff
@@ -134,3 +140,85 @@ func TestHandleUnexpectedError(t *testing.T) {
...
+		}
+
+		w := httptest.NewRecorder()
+
+		api.Router.ServeHTTP(w, r)
+
+		So(w.Code, ShouldEqual, http.StatusInternalServerError)
+	})
+
+	Convey("Global Sign Out returns 400: request error", t, func() {
```
mike-kelly-methods, line 212: Could combine all `400` tests by putting all test scenarios in a struct and iterating over them using:

`for _, tt := range <STRUCT_NAME> { ... }`

api/tokens.go
mwnicks: I think thats the right approach in general.
But in this case all the validation checks are on the same piece of input and if one fails all subsequent ones can't pass, and in some cases would cause panic states (e.g. trying to access the 2nd element after the split when the header was malformed and doesn't have a second element).
So I felt it was better to return after a validation failure rather than get in to a lot of nested ifs.

mike-kelly-methods: 👍

mike-kelly-methods (approved):
👍

PR #22, opened 2021-05-13
Feature/update login endpoint with cognito

HeJones:
### What

- Call InititateAuth endpoint in Cognito
- Return 201 with tokens in the header and expiration time in the body if authorised
- Return 401 if not authorised
- Return 403 if too many attempts
- Return 500 if internal server error 
- Return 400 for other client errors from cognito

- Add unit tests
- Add feature tests and cognito stub
### How to review

- make test
- make test-component
- Run make debug and a post request can be made to /tokens with a user to see the different scenarios. Please let me know fi you need user details

### Who can review

Anyone but @AWitcherONS and myself


api/tokens.go
```diff
@@ -165,3 +219,74 @@ func handleUnexpectedError(ctx context.Context, w http.ResponseWriter, err error
 	writeErrorResponse(ctx, w, http.StatusInternalServerError, errorResponseBody)
 	return
 }
+
+func computeSecretHash(clientSecret string, username string, clientId string) string {
```
mwnicks: This may want to be refactored to a more shared location at some point as other endpoints will need it

api/tokens.go
```diff
@@ -165,3 +219,74 @@ func handleUnexpectedError(ctx context.Context, w http.ResponseWriter, err error
...
+		ClientMetadata:    map[string]*string{},
+		UserContextData:   &cognitoidentityprovider.UserContextDataType{},
+	}
+
+	return authInput
+}
+
+func buildSucessfulResponse(result *cognitoidentityprovider.InitiateAuthOutput, w http.ResponseWriter, ctx context.Context) {
+
+	if result.AuthenticationResult != nil {
```
mwnicks, line 245: If this condition is not met the endpoint will return an empty 200 response.
Probably a good idea to have an else to catch anything unexpected from Cognito and return a 500

api/tokens_test.go
```diff
@@ -139,12 +141,109 @@ func TestHandleUnexpectedError(t *testing.T) {
 		So(resp.Body.String(), ShouldResemble, errorResponseBodyExample)
 	})
 }
+func TestCognitoResponseBuild(t *testing.T) {
```
mwnicks: Slightly confusing name, should it not be TestCognitoRequestBuild?

api/tokens_test.go
```diff
@@ -139,12 +141,109 @@ func TestHandleUnexpectedError(t *testing.T) {
...
+
+		So(*response.AuthParameters["USERNAME"], ShouldEqual, authParams.Email)
+		So(*response.AuthParameters["PASSWORD"], ShouldEqual, authParams.Password)
+		So(*response.AuthParameters["SECRET_HASH"], ShouldNotBeEmpty)
+		So(*response.AuthFlow, ShouldResemble, "authflow")
+		So(*response.ClientId, ShouldResemble, "awsclientid")
+	})
+}
+
+func TestCognitoRequestHeaderBuild(t *testing.T) {
```
mwnicks: Similar to previous test, should this be TestCognitoResponseHeaderBuild?

mwnicks (changes requested):
Should the 'main' file have been committed?

api/tokens_test.go
HeJones: This has been updated to the suggestion

api/tokens_test.go
HeJones: This has been updated to the suggestion

api/tokens.go
HeJones: Yes, well spotted. Thank you for the suggestion

api/tokens.go
HeJones: Yes good idea, we have created a utility package

mwnicks (approved):

PR #26, opened 2021-05-17
refactor to handle all required Cognito errors as internal errors

mwnicks:
### What

Refactored sign out endpoint to use shared error checking for internal errors

### How to review

Ensure the unit and feature tests are still passing

### Who can review

Anyone but me


HeJones (approved):
It looks good to me, and the tests pass locally

PR #28, opened 2021-05-20
Feature/refresh tokens

mwnicks:
### What

Added an endpoint to refresh a users access token 

### How to review

Check the unit and feature tests are passing.

Use the localhost:25600/tokens endpoint to generate session tokens.
Add the generated ID and refresh tokens to a PUT request to localhost:25600/tokens/self
You should receive a 201 response with a new Access token and expiry time

### Who can review

Anyone but me


api/api.go
```diff
@@ -40,6 +40,7 @@ func Setup(ctx context.Context, r *mux.Router, cognitoClient cognito.Client, use
 
 	r.HandleFunc("/tokens", api.TokensHandler(ctx)).Methods("POST")
 	r.HandleFunc("/tokens/self", api.SignOutHandler(ctx)).Methods("DELETE")
+	r.HandleFunc("/tokens/self", api.RefreshHandler(ctx)).Methods("PUT")
```
eldeal, line 44: As someone not familiar with this service, /tokens/self strikes me as an unusual departure from RESTful routes. It might be worth adding a comment here to explain why we're not using some identifier in place of self?

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
...
 }
 
+func (api *API) RefreshHandler(ctx context.Context) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		var errorList []apierrorsdeprecated.Error
+
+		refreshToken := models.RefreshToken{
+			TokenString: req.Header.Get(RefreshTokenHeaderName),
+		}
+		refreshToken.Validate(ctx, &errorList)
```
eldeal: Validate methods would probably read better if they returned []apierrors.Error, then you can choose whether to append them or just len(list) != 0 depending on where its used.

so this line could become `errorList = append(errorList, refreshToken.Validate)` and if there are no errors the list should still be empty

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
 	}
 }
 
+func (api *API) RefreshHandler(ctx context.Context) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		var errorList []apierrorsdeprecated.Error
```
eldeal, line 178: Consider moving all of this token validation/error handling to a separate function - it seems like the kind of thing that might happen in several handlers?

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
...
+			errorResponseBody := apierrorsdeprecated.ErrorResponseBodyBuilder(errorList)
+			apierrorsdeprecated.WriteErrorResponse(ctx, w, http.StatusBadRequest, errorResponseBody)
+			return
+		}
+
+		authInput := refreshToken.GenerateRefreshRequest(api.ClientSecret, idToken.Claims.CognitoUser, api.ClientId)
+		result, authErr := api.CognitoClient.InitiateAuth(authInput)
+
+		if authErr != nil {
+			log.Event(ctx, authErr.Error(), log.ERROR)
```
eldeal: This log event would benefit from a message indicating the context where the authErr was encountered. If the Cognito Client is called from multiple places, knowing that this specifically happened in the refresh handler would be helpful

apierrorsdeprecated/apierrorsdeprecated.go
```diff
@@ -46,19 +54,19 @@ var ErrInvalidEmail = errors.New("invalid email")
 
 var ErrDuplicateEmail = errors.New("duplicate email")
 
-func IndividualErrorBuilder(err error, description string) (Error errModels.Error) {
+func IndividualErrorBuilder(err error, description string) (apiErr Error) {
 
-	Error = errModels.Error{
+	apiErr = Error{
```
eldeal, line 59: can just do `return Error{ ` and save the named variable and the mem allocation

apierrorsdeprecated/apierrorsdeprecated.go
```diff
@@ -46,19 +54,19 @@ var ErrInvalidEmail = errors.New("invalid email")
...
 
-	return Error
+	return apiErr
 }
 
-func ErrorResponseBodyBuilder(listOfErrors []errModels.Error) (errorResponseBody errModels.ErrorList) {
+func ErrorResponseBodyBuilder(listOfErrors []Error) (errorResponseBody ErrorList) {
 
-	errorResponseBody = errModels.ErrorList{
+	errorResponseBody = ErrorList{
```
eldeal, line 69: same as above, return ErrorList{

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
...
+			if authErr.Error() == "NotAuthorizedException: Refresh Token has expired" {
+				expiredTokenError := apierrorsdeprecated.IndividualErrorBuilder(authErr, apierrorsdeprecated.TokenExpiredMessage)
+				errorList = append(errorList, expiredTokenError)
+				errorResponseBody := apierrorsdeprecated.ErrorResponseBodyBuilder(errorList)
```
eldeal, line 204: errorList at this point can only ever have 1 element in it. rather than doing the append on the previous line, you might be able to use variadic notation to pass your singular error in: `ErrorResponseBodyBuilder(expiredTokenError...)` 

Even if you can't do this at the moment because of the signature:
`ErrorResponseBodyBuilder(listOfErrors []Error)`
if you changed it to `ErrorResponseBodyBuilder(listOfErrors Error...)` it would allow you to pass in 1 or many items of the same type, and they'll be treated as an array within the function. Might clean up your usages a bit

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
...
+				apierrorsdeprecated.WriteErrorResponse(ctx, w, http.StatusForbidden, errorResponseBody)
+				return
+			} else {
+				apierrorsdeprecated.HandleUnexpectedError(ctx, w, authErr, apierrorsdeprecated.InternalErrorMessage)
+				return
```
eldeal: Very minor, I'd personally put the return as the last line in the containing if statement, rather than 2 separate occurences.

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
 	}
 }
 
+func (api *API) RefreshHandler(ctx context.Context) http.HandlerFunc {
```
eldeal, line 176: Exported functions should have comments please

api/tokens.go
```diff
@@ -162,6 +170,51 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
...
+				errorResponseBody := apierrorsdeprecated.ErrorResponseBodyBuilder(errorList)
+				apierrorsdeprecated.WriteErrorResponse(ctx, w, http.StatusForbidden, errorResponseBody)
+				return
+			} else {
+				apierrorsdeprecated.HandleUnexpectedError(ctx, w, authErr, apierrorsdeprecated.InternalErrorMessage)
+				return
+			}
+		}
+
+		result.AuthenticationResult.RefreshToken = &refreshToken.TokenString
```
eldeal: It seems weird/non-obvious to me why when we're calling a refresh endpoint we can assign our input value as the result. Like shouldn't we be getting a new token? But this is likely because I'm uninformed on the topic, and might be where the method having some godoc as mentioned above would help. But just triple checking this is the correct behaviour? :)

apierrorsdeprecated/errorModels.go
```diff
@@ -1,4 +1,4 @@
-package models
+package apierrorsdeprecated
```
eldeal, line 1: Why choose the name deprecated? I think we have apierrors packages in other services, so having the errors in their own package I'm totally on board with, just not sure on the naming :)

api/tokens_test.go
```diff
@@ -331,3 +333,126 @@ func TestSignOutHandler(t *testing.T) {
...
+		r                                                     = mux.NewRouter()
+		ctx                                                   = context.Background()
+		poolId, clientId, clientSecret, clientAuthFlow string = "us-west-11_bxushuds", "client-aaa-bbb", "secret-ccc-ddd", "authflow"
+	)
+
+	m := &mock.MockCognitoIdentityProviderClient{}
+
+	api, _ := Setup(ctx, r, m, poolId, clientId, clientSecret, clientAuthFlow)
+
+	Convey("Refresh returns 201: successfully refreshed access token", t, func() {
```
eldeal: These unit tests use a real mux.NewRouter to call the API and check we're getting the right status codes. This seems to be exactly the same thing we're testing in the feature tests. I dont think we need them both? Might be one to discuss within the team, maybe I have the wrong end of the stick, but I would suggest dropping the unit tests for places HTTP handler response options

models/tokens_models.go
```diff
@@ -0,0 +1,98 @@
...
+	TokenString string
+	Claims      IdClaims
+}
+
+func (t *IdToken) ParseWithoutValidating(tokenString string) error {
+	parser := new(jwt.Parser)
+
+	idToken, _, parsingErr := parser.ParseUnverified(tokenString, &IdClaims{})
+	if parsingErr != nil {
+		fmt.Println(parsingErr)
```
eldeal: if we're going to log it should use our logger. Possible this is just an old debug statement you want to remove? likely that error is getting logged where its handled, so logging here may be unnecesary

models/tokens_models.go
```diff
@@ -0,0 +1,98 @@
...
+	return nil
+}
+
+func (t *IdToken) Validate(ctx context.Context, errorList *[]apierrorsdeprecated.Error) {
+	if t.TokenString == "" {
+		invalidIDTokenErrorBody := apierrorsdeprecated.IndividualErrorBuilder(apierrorsdeprecated.InvalidIDTokenError,
+			apierrorsdeprecated.MissingIDTokenMessage)
+		*errorList = append(*errorList, invalidIDTokenErrorBody)
+		log.Event(ctx, apierrorsdeprecated.MissingRefreshTokenMessage, log.ERROR)
+	} else {
```
eldeal: We can avoid the else if we return from the end of the 'if' block. Then all the contained code moves back to the primary line of sight / happy path

models/tokens_models.go
```diff
@@ -0,0 +1,98 @@
...
+	authInput := &cognitoidentityprovider.InitiateAuthInput{
+		AuthFlow:       &refreshAuthFlow,
+		AuthParameters: authParams,
+		ClientId:       &clientId,
+	}
+	return authInput
+}
+
+func (t *RefreshToken) Validate(ctx context.Context, errorList *[]apierrorsdeprecated.Error) {
+	if t.TokenString == "" {
```
eldeal: For primary line of sight reasons, if you did `if t.TokenString != "" { return }` then all the rest of the code can come back to the happy path

eldeal (changes requested):

apierrorsdeprecated/errorModels.go
mwnicks: We have tickets to refactor the error handling across the entire service. The first of these was done whilst I was working on this feature, so this is me fixing merge conflicts.
The renaming of the package to deprecated is to support the existing endpoints whilst they are being migrated to the new error handling.

apierrorsdeprecated/apierrorsdeprecated.go
mwnicks: The error handling is being refactored in a separate ticket, as such are you happy for this to be left as is until this package is removed?

apierrorsdeprecated/apierrorsdeprecated.go
mwnicks: The error handling is being refactored in a separate ticket, as such are you happy for this to be left as is until this package is removed?

models/tokens_models.go
mwnicks: Yeah this is debug statement I meant to remove. 
I will delete it now

api/tokens.go
mwnicks: Yeah that is cleaner

api/api.go
mwnicks: I've added a comment to explain this. The identifier for these endpoint are the JWT tokens that are passed in the request headers.

api/tokens.go
mwnicks: I've updated the Validate methods to return errorlist.

api/tokens.go
mwnicks: The error handling is being refactored in a separate ticket, as such are you happy for this to be left as is in this branch to refactored in that ticket?

api/tokens.go
mwnicks: Added context about the Cognito request being made and where in the identity-api the call is being made.

api/tokens.go
mwnicks: The error handling is being refactored in a separate ticket, as such are you happy for this to be left as is in this branch to refactored in that ticket?

api/tokens.go
mwnicks: Comments added to the exported functions

api/tokens.go
mwnicks: The refresh token is used to create new access and Id tokens to keep the session alive. A new refresh token is not generated, a new refresh token can only be generated by the user logging in again.
But thinking about it, the client already has the refresh token so it doesn't need to be returned again.

api/tokens_test.go
mwnicks: I was thinking this as I was writing them, I've taken them out now as I think you are right.

models/tokens_models.go
mwnicks: Updated to as requested

models/tokens_models.go
mwnicks: Updated to as requested

api/tokens.go
eldeal: Ah I also meant that because you can just return the errors you need you dont need to pass errorList in at all. Hence you'd need to append to the error list in the scope of your handler, because your refreshToken.Validate only returns the errors for itself (and in fact could just return singular instead of [] I think). 

This might come under the error handling work youve mentioned though.

eldeal (approved):

PR #30, opened 2021-05-21
Feature/update login endpoint

HeJones:
### What

- Add in AdminUserGlobalSignOut call before login
- Add in part one of the changes for work to change how we handle errrors
### How to review

- Run `make test` & `make test-component`
### Who can review

Anyone but me


api/tokens.go
```diff
@@ -46,6 +47,26 @@ func (api *API) TokensHandler(ctx context.Context) http.HandlerFunc {
 		validEmailRequest := emailValidation(authParams)
 
 		if validPasswordRequest && validEmailRequest {
+
+			err = adminUserGlobalSignOut(authParams, api.UserPoolId, api.CognitoClient)
```
daiLlew: Consider renaming this - something like - `terminateExistingSession` is a bit more user friendly and a bit more obvious about what its doing.

api/tokens.go
```diff
@@ -162,6 +183,15 @@ func (api *API) SignOutHandler(ctx context.Context) http.HandlerFunc {
 	}
 }
 
+func adminUserGlobalSignOut(authParams AuthParams, userPoolId string, cognitoClient cognito.Client) (err error) {
+	adminUserGlobalSignOutInput := buildAdminSignoutRequest(authParams, userPoolId)
+	_, err = cognitoClient.AdminUserGlobalSignOut(adminUserGlobalSignOutInput)
```
daiLlew, line 246: Is err the only indicator this action failed? For example executing the http request was successfully but the response returned contained a non success status code. Does that return an error or do you have to check the other return var to check if it was successful?

api/tokens.go
```diff
@@ -197,6 +227,16 @@ func buildCognitoRequest(authParams AuthParams, clientId string, clientSecret st
...
 }
 
+func buildAdminSignoutRequest(authParams AuthParams, userPoolId string) (adminSignoutInput *cognitoidentityprovider.AdminUserGlobalSignOutInput) {
+
+	adminSignoutInput = &cognitoidentityprovider.AdminUserGlobalSignOutInput{
+		Username:   &authParams.Email,
+		UserPoolId: &userPoolId,
+	}
+
+	return adminSignoutInput
```
daiLlew: You may as well move this into the  `adminUserGlobalSignOut` func - it would use less lines.

api/tokens.go
HeJones: Thank you, this is a good idea

api/tokens.go
HeJones: This is a really good question, because it wasn't too clear to me from the documentation: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminUserGlobalSignOut.html#API_AdminUserGlobalSignOut_ResponseElements

If possible itt would be really good to have chat about this

api/tokens.go
HeJones: Thank you

api/tokens.go
daiLlew: The doc isn't clear to me but I think its saying that it will return an `awserr.Error` for all failures. So what you have here is fine based on that assumption/understanding.

daiLlew (dismissed):

daiLlew (approved):

PR #31, opened 2021-05-24
Feature/api errors refine

mwnicks:
### What

Refactored the error handling across the service to improve consistency and reduce repetition

### How to review

Ensure all the unit and feature tests are still passing

### Who can review

Anyone but me


models/errors.go
```diff
@@ -0,0 +1,79 @@
+package models
+
+import (
+	"context"
+	"errors"
+	"github.com/ONSdigital/log.go/log"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+)
+
+type ErrorList struct {
```
daiLlew: I think something like `ErrorResponse` or `ErrorEntity` or `ErrorBody` would be a better name as its not strictly a list - it contains a list. But that is me being very nit picky.

models/errors.go
```diff
@@ -0,0 +1,79 @@
...
+	log.Event(ctx, description, log.Error(cause), log.ERROR)
+	return &Error{
+		Cause:       cause,
+		Code:        code,
+		Description: description,
+	}
+}
+
+func NewValidationError(ctx context.Context, code string, description string) *Error {
+	log.Event(ctx, code+": "+description, log.ERROR)
```
daiLlew: Minor point but our logging standards state we should avoid building dynamic strings for log events - we want the event string to be the same for every occurrence of a specific event. Can you update this to add the code the `log.Data` please? That way the event is the same regardless of the code and the code is additional data about the event.
````
log.Event(ctx, description, log.ERROR, log.Data{"code": code})
````

File: models/errors.go [hunk 3]
```diff
@@ -0,0 +1,79 @@
...
+
+func (e *CognitoError) Error() string {
+	if e.Cause != nil {
+		return e.Cause.Error()
+	}
+	return e.Code + ": " + e.Description
+}
+
+func NewCognitoError(ctx context.Context, err error, errContext string) CognitoError {
+	log.Event(ctx, errContext+" - "+err.Error(), log.ERROR)
```
daiLlew: Same here with the event string:

```
log.Event(ctx, errContext, log.Error(err), log.ERROR)
```
The context is the event and the error is meta data about the event.

models/errors.go
```diff
@@ -0,0 +1,79 @@
...
+	cognitoErr := err.(awserr.Error)
```
daiLlew: Do you need to check this cast? If so consider using `err.As`/`err.Is` (which ever is appropriate).

models/errors.go
```diff
@@ -0,0 +1,79 @@
...
+		Code:        code,
+		Description: cognitoErr.Message(),
+	}
+}
+
+func mapCognitoErrorToLocalError(cognitoErr awserr.Error) string {
+	if val, ok := cognitoErrorMapping[cognitoErr.Code()]; ok {
+		return val
+	}
+	return InternalError
```
daiLlew, line 70: Probably worth adding a log event here to say defaulting to Internal server error - just for visibility of whats happening.

daiLlew (commented):
I've added a few very minor comments to consider but overall I think this is a neat design that solves the problem so I'm happy to +1 it.

I won't approve yet in case anyone else whats to add any comments or questions.

models/errors.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/ONSdigital/log.go/log"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+)
+
+type ErrorList struct {
+	Errors []BaseError `json:"errors"`
+	Status int         `json:"-"`
+}
+
+type BaseError interface {
```
ChrysmOre: Why is this needed? This is exactly what a standard Go error is (an interface that exposes func Error() string)

models/errors.go
ChrysmOre: I think it's good practice to use FooResponse and FooRequest whenever that's what the object represents, here is a good example.

models/errors.go
```diff
@@ -0,0 +1,79 @@
...
+func NewValidationError(ctx context.Context, code string, description string) *Error {
+	log.Event(ctx, code+": "+description, log.ERROR)
+	return &Error{
+		Cause:       errors.New(code),
+		Code:        code,
+		Description: description,
+	}
+}
+
+type CognitoError struct {
```
ChrysmOre: Is this separate type actually necessary? It seems to be identical to the standard Error type you have, the only thing being different is how it is constructed with NewCognitoError.. could you not just return a regular Error from that func as you do with NewValidationError?

api/tokens.go
```diff
@@ -124,52 +123,35 @@ func (api *API) TokensHandler(ctx context.Context) http.HandlerFunc {
...
-			errorResponseBody := apierrorsdeprecated.ErrorResponseBodyBuilder(errorList)
-			apierrorsdeprecated.WriteErrorResponse(ctx, w, http.StatusBadRequest, errorResponseBody)
-			return
-		}
+func (api *API) SignOutHandler(w http.ResponseWriter, req *http.Request, ctx context.Context, errorResponse *models.ErrorResponse) {
+	accessToken := models.AccessToken{
+		AuthHeader: req.Header.Get(AccessTokenHeaderName),
+	}
+	validationErr := accessToken.Validate(ctx)
+	if validationErr != nil {
```
eldeal, line 152: I think there would be some value in simplifying a step further, towards something like:

```suggestion
	if validationErr != nil {
	    return ErrorResponse{
	        Errors: validationErr,
	        Status: http.StatusBadRequest
	        }
	   }
```
- As soon as we encounter an error we should look to return, so theres no need to wait and do the len(errors) check in a separate step
- Since our handlers are all wrapped in baseHandler, we can actually return errors from here, rather than have to pass the additional errorResponse argument in - save the memory allocation until its actually needed.

api/tokens.go
```diff
@@ -124,52 +123,35 @@ func (api *API) TokensHandler(ctx context.Context) http.HandlerFunc {
...
-			isInternalError := apierrorsdeprecated.IdentifyInternalError(err)
-			if isInternalError {
-				apierrorsdeprecated.WriteErrorResponse(ctx, w, http.StatusInternalServerError, errorResponseBody)
-			} else {
-				apierrorsdeprecated.WriteErrorResponse(ctx, w, http.StatusBadRequest, errorResponseBody)
-			}
-			return
+	if err != nil {
+		responseErr := models.NewCognitoError(ctx, err, "Cognito GlobalSignOut request for signout")
+		errorResponse.Errors = append(errorResponse.Errors, &responseErr)
```
eldeal: If we make the change mentioned above, not passing in an errorResponse struct to this handler, then you don't need to append here you can just instantiate. The benefit of this is it gives less impression that there might already be a few errors waiting to be processed; we're more clearly saying "This function I just called might have a couple of errors, but I know I've handled all errors before this point"

models/errors.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/ONSdigital/log.go/log"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+)
+
+type ErrorResponse struct {
+	Errors []BaseError `json:"errors"`
+	Status int         `json:"-"`
+}
+
+type BaseError interface {
```
eldeal: II think error is already an interface in go, so I'm not actually sure this interface type adds value. On L32, you could just return `error` as the type and other than it being a pointer I think your custom types would be valid

ChrysmOre (changes requested):

HeJones (dismissed):
It looks good to me

models/errors.go
mwnicks: I've removed this interface to directly use the error interface instead.

api/tokens.go
mwnicks: Changed to return a new ErrorResponse when an error is created.

api/tokens.go
mwnicks: Changed to return a new ErrorResponse when an error is created.

models/errors.go
mwnicks: I've removed this interface to directly use the error interface instead.

models/errors.go
mwnicks: Agreed, it started out with different requirements but they are now the same.
I've removed this type and the constructor now returns the regular Error type.

api/api.go
```diff
@@ -20,6 +23,18 @@ type API struct {
 	ClientAuthFlow string
 }
 
+type baseHandler func(w http.ResponseWriter, r *http.Request, ctx context.Context) *models.ErrorResponse
+
+func (handler baseHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
eldeal: I wonder if this would be a little easier (and shorter on the routing declaration) if we treated it more as middleware, e.g.:

``` go
func contextAndErrors(h baseHandler) http.HandlerFunc {
	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {

		ctx := req.Context()

		if err := h(w, req, ctx); err != nil {
			WriteErrorResponse(ctx, w, err)
		}
	})
}

...
r.HandleFunc("/tokens/self", contextAndErrors(api.SignOutHandler)).Methods("DELETE")
```

Then we don't need the ServeHTTP thing on the routes. Better names may be available.

eldeal: This function has also removed the double-instantiation of ctx, and shortens the err var name - both mild suggestions but take or leave

api/api.go
```diff
@@ -40,8 +55,28 @@ func Setup(ctx context.Context, r *mux.Router, cognitoClient cognito.Client, use
...
 	r.HandleFunc("/tokens", api.TokensHandler(ctx)).Methods("POST")
 	// self used in paths rather than identifier as the identifier is JWT tokens passed in the request headers
-	r.HandleFunc("/tokens/self", api.SignOutHandler(ctx)).Methods("DELETE")
+	r.HandleFunc("/tokens/self", baseHandler(api.SignOutHandler).ServeHTTP).Methods("DELETE")
 	r.HandleFunc("/tokens/self", api.RefreshHandler(ctx)).Methods("PUT")
 	r.HandleFunc("/users", api.CreateUserHandler(ctx)).Methods("POST")
 	return api, nil
 }
+
+func WriteErrorResponse(ctx context.Context, w http.ResponseWriter, errorResponse *models.ErrorResponse) {
```
eldeal: I haven't looked closely, but does this function actually need to be exported?

mwnicks: No it doesn't, that is leftover from where it used to be in a different package.
I'll update it

api/api.go
mwnicks: I didn't like how the route registration looked, this is much nice

mwnicks: I think the double-instantiation of ctx was a mistake anyway

eldeal (approved):

ChrysmOre (approved):

PR #33, opened 2021-05-27
Feature/sign in new errors

mwnicks:
### What

Refactor sign in endpoint to implement new error handling pattern

### How to review

Ensure the unit and feature tests are still passing

### Who can review

Anyone but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -37,6 +37,7 @@ func contextAndErrors(h baseHandler) http.HandlerFunc {
 		response, err := h(w, req, ctx)
 		if err != nil {
 			writeErrorResponse(ctx, w, err)
+			return
 		}
 		writeSuccessResponse(ctx, w, response)
 	}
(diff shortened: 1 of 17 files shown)


api/api.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/gorilla/mux"
 )
 
+var (
+	IdTokenHeaderName      = "ID"
+	AccessTokenHeaderName  = "Authorization"
+	RefreshTokenHeaderName = "Refresh"
+)
```
mike-kelly-methods, line 20: Can't see these variables being used anywhere?

api/api.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/gorilla/mux"
 )
 
+var (
```
mike-kelly-methods, line 16: As this `var()` block is only used by the `Tokens` package - it could be defined within that package to reduce scope. Also, would be better as a `map[string]string`?

mike-kelly-methods (approved):
👍

PR #37, opened 2021-06-03
Bug/login race condition

mwnicks:
### What

Removed the sign out old sessions step of sign in handler.
[Ticket](https://trello.com/c/zgjCdOhd/504-race-condition-in-signing-out-a-users-existing-sessions-before-signing-them-in)

### How to review

Ensure all unit and feature tests are still passing.

### Who can review

Anyone but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -26,12 +26,12 @@ type API struct {
 	ClientAuthFlow string
 }
 
-type baseHandler func(w http.ResponseWriter, r *http.Request, ctx context.Context) (*models.SuccessResponse, *models.ErrorResponse)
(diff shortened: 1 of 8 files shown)


cadmiumcat (commented):
LGTM and tests pass. Altough I don't undestand enough to get why the scenarios were removed - I assume they dont' apply anymore?
I'm so glad you moved the `ctx`!!!!! 😆

mwnicks:
> LGTM and tests pass. Altough I don't undestand enough to get why the scenarios were removed - I assume they dont' apply anymore?

Correct, the scenarios removed related to expected behaviour depending on different outcomes from the handler step that has been removed. The remaining scenarios cover the expected behaviour of the remaining code.

cadmiumcat (approved):
I'm happy with this :)

PR #40, opened 2021-06-15
Add NEW_PASSWORD_REQUIRED challenge response in Tokens handler.

mike-kelly-methods:
### What

As per title. Add challenge response handling logic so that correct response data can be returned to caller.

### How to review

Review code, ensure both unit and feature tests are passing.

### Who can review

Anyone but me.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -3,8 +3,9 @@ package api
 import (
 	"context"
 	"encoding/json"
-	"github.com/ONSdigital/dp-identity-api/models"
 	"net/http"
+
+	"github.com/ONSdigital/dp-identity-api/models"
 
 	"github.com/ONSdigital/dp-identity-api/cognito"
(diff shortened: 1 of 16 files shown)


mwnicks (dismissed):

api/api.go
```diff
@@ -19,6 +19,7 @@ var (
 	ONSRealm               = "Florence publishing platform"
 	Charset                = "UTF-8"
 	NewPasswordChallenge   = "NEW_PASSWORD_REQUIRED"
+	MFASetupChallenge      = "MFA_SETUP"
```
mwnicks: This appears to be an unused variable.

mike-kelly-methods: Yeah, James mentioned that we should think about catering for MFA challenges too. I can take it out for now. Include in a new ticket.

mwnicks (approved):

PR #41, opened 2021-06-17
Feature/change password

mwnicks:
### What

Added the endpoint to handle password change requests.

### How to review

Ensure all tests are passing.

1. Create a new users using the /v1/users POST endpoint.
2. Attempt to log in against the /v1/tokens POST endpoint using the credentials emailed to you after user creation 
3. Send a request to the /v1/users/self/password PUT endpoint - the body of the request should contain:
    - "type": "NewPasswordRequired"
    - "email": the email of the user you created in step 1
    - "password": a new password (must contain 1 uppercase, 1 lowercase, 1 numeric and be longer than 14 characters)
    - "session": returned in the response in step 2
4. You should receive a 202 response with the 'Authorization', 'ID' and 'Refresh' headers set.

### Who can review

Anyone but me


HeJones:
I have been working through the steps to review, but I can't complete step 2 because the username isn't a valid email address, however I also get the same error when I try to use the email address

HeJones (approved):
The changes look good to me and I was able to follow the steps and get a 202 once my account had been manually verified

PR #42, opened 2021-06-21
Feature/request password reset

mwnicks:
### What

Added endpoint to handle requests for password reset emails 

### How to review

Ensure the unit and feature tests are passing.

*With a user created and confirmed within Cognito
- Send a POST request to '/v1/password-reset' - the body of the request should contain:
    - 'email': the email of the user
- You should receive a 202 response and an email with a verification code

- Send a POST request to '/v1/password-reset' - the body of the request should contain:
    - 'email': an email that does not relate to a user in Cognito
- You should receive a 202 response but no email

### Who can review

Anyone but me

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -151,17 +151,10 @@ func (api *API) PasswordResetHandler(ctx context.Context, w http.ResponseWriter,
 
 	forgotPasswordRequest := passwordResetParams.BuildCognitoRequest(api.ClientSecret, api.ClientId)
 
-	result, err := api.CognitoClient.ForgotPassword(forgotPasswordRequest)
+	_, err = api.CognitoClient.ForgotPassword(forgotPasswordRequest)
 	if err != nil {
 		responseErr := models.NewCognitoError(ctx, err, "ForgotPassword request from password reset endpoint")
 		if responseErr.Code == models.InternalError {
-			return nil, models.NewErrorResponse([]error{responseErr}, http.StatusInternalServerError, nil)
-		}
-	}
-
-	if result != nil {
-		responseErr := passwordResetParams.BuildSuccessfulJsonResponse(ctx, result)
-		if responseErr != nil {
 			return nil, models.NewErrorResponse([]error{responseErr}, http.StatusInternalServerError, nil)
 		}
 	}
--- a/api/users.go
+++ b/api/users.go
@@ -127,3 +127,44 @@ func (api *API) ChangePasswordHandler(ctx context.Context, w http.ResponseWriter
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusAccepted, headers), nil
 }
+
+//PasswordResetHandler creates a new user and returns a http handler interface
+func (api *API) PasswordResetHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	defer req.Body.Close()
+
+	body, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		return nil, handleBodyReadError(ctx, err)
+	}
+
+	passwordResetParams := models.PasswordReset{}
+	err = json.Unmarshal(body, &passwordResetParams)
+	if err != nil {
+		return nil, handleBodyUnmarshalError(ctx, err)
+	}
+
+	validationErr := passwordResetParams.Validate(ctx)
+
+	if validationErr != nil {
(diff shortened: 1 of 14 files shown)


api/users.go
```diff
@@ -127,3 +127,44 @@ func (api *API) ChangePasswordHandler(ctx context.Context, w http.ResponseWriter
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusAccepted, headers), nil
 }
+
+//PasswordResetHandler creates a new user and returns a http handler interface
```
sandypadmanabhan: comment to be updated

api/users.go
```diff
@@ -127,3 +127,44 @@ func (api *API) ChangePasswordHandler(ctx context.Context, w http.ResponseWriter
...
+	if validationErr != nil {
+		return nil, models.NewErrorResponse([]error{validationErr}, http.StatusBadRequest, nil)
+	}
+
+	forgotPasswordRequest := passwordResetParams.BuildCognitoRequest(api.ClientSecret, api.ClientId)
+
+	result, err := api.CognitoClient.ForgotPassword(forgotPasswordRequest)
+	if err != nil {
+		responseErr := models.NewCognitoError(ctx, err, "ForgotPassword request from password reset endpoint")
+		if responseErr.Code == models.InternalError {
```
sandypadmanabhan: cognito response has 2 error code 400 and 500, but we are checking for just one here

api/users_test.go
```diff
@@ -378,3 +379,115 @@ func TestChangePasswordHandler(t *testing.T) {
...
+					awsOrigErr := errors.New(awsErrCode)
+					awsErr := awserr.New(awsErrCode, awsErrMessage, awsOrigErr)
+					return nil, awsErr
+				},
+				http.StatusInternalServerError,
+			},
+			{
+				// Cognito invalid user
+				func(input *cognitoidentityprovider.ForgotPasswordInput) (*cognitoidentityprovider.ForgotPasswordOutput, error) {
+					awsErrCode := "CodeMismatchException"
```
sandypadmanabhan: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ForgotPassword.html- Cannot find "CodeMismatchException". Am I looking at the correct list

api/users.go
mwnicks: Updated to correctly describe the function

api/users_test.go
mwnicks: I change the test scenario but missed changing the AWS Error code.
Should be 'UserNotFoundException', I've updated it now.
Tests were passing as they are both 400 errors handled in the same way in the handler.

sandypadmanabhan (commented):
Do we want to update the swagger as well?

api/users.go
```diff
@@ -127,3 +127,39 @@ func (api *API) ChangePasswordHandler(ctx context.Context, w http.ResponseWriter
...
 	return models.NewSuccessResponse(jsonResponse, http.StatusAccepted, headers), nil
 }
+
+//PasswordResetHandler requests a password reset email be sent to the user and returns a http handler interface
+func (api *API) PasswordResetHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	defer req.Body.Close()
+
+	body, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		return nil, handleBodyReadError(ctx, err)
```
sandypadmanabhan, line 137: We might need to log the event when an error is returned unless we have decided not to...

mwnicks: The new error constructors write a log entry for the error when the error is built.
See here https://github.com/ONSdigital/dp-identity-api/blob/develop/models/errors.go#L23

swagger.yaml
```diff
@@ -223,6 +223,38 @@ paths:
...
+        202:
+          description: "Request accepted"
+        400:
+          description: "Email was not in a valid format"
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        500:
+          description: "Invalid request body or unexpected Cognito response"
+          schema:
+            $ref: '#/definitions/ErrorResponse'
```
sandypadmanabhan, line 258: can we add usernotfound error to the error object list

mwnicks: Good spot, I've added it now

sandypadmanabhan (approved):
LGTM

PR #44, opened 2021-06-24
Feature/list users

mwnicks:
### What

Added an endpoint to list the users that are currently in the user pool

### How to review

Check all the unit and feature tests are passing
Call GET /v1/users you should receive a 200 response with JSON containing a list of users and a count of the number of users in the list.

### Who can review

Anyone but me


DavidSubiros (commented):
- The code looks good to me.
- Unit tests pass
- Component tests pass
- Verified that `GET http://localhost:25600/v1/users` returns a list of users coming from Cognito
- Modified my aws credentials, run the service again, and then it fails to run, which is expected.

DavidSubiros (approved):

PR #51, opened 2021-07-05
Feature/update user

mwnicks:
### What

Added the update user endpoint

### How to review

Ensure all unit and component tests are passing.
Send a request to PUT '/v1/users/{user_id}' with user_id = ae54bd71-6099-4364-acd3-519a64202679 and request body:
- forename = a new forename for the user
- lastname = a new surname for the user
- status = CONFIRMED

You should get a 200 response with the response body like:
{
       "id": "ae54bd71-6099-4364-acd3-519a64202679",
       "forename": <new forename>,
       "lastname": <new lastname>,
       "email": "matt.nicks@ons.gov.uk",
       "groups": [],
       "status": "CONFIRMED"
}

### Who can review

Any one but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -66,18 +66,18 @@ func Setup(ctx context.Context, r *mux.Router, cognitoClient cognito.Client, use
 		ClientAuthFlow: clientAuthFlow,
 	}
 
-	r.HandleFunc("/v1/tokens", contextAndErrors(api.TokensHandler)).Methods("POST")
+	r.HandleFunc("/v1/tokens", contextAndErrors(api.TokensHandler)).Methods(http.MethodPost)
 	// self used in paths rather than identifier as the identifier is JWT tokens passed in the request headers
-	r.HandleFunc("/v1/tokens/self", contextAndErrors(api.SignOutHandler)).Methods("DELETE")
-	r.HandleFunc("/v1/tokens/self", contextAndErrors(api.RefreshHandler)).Methods("PUT")
-	r.HandleFunc("/v1/users", contextAndErrors(api.CreateUserHandler)).Methods("POST")
-	r.HandleFunc("/v1/users", contextAndErrors(api.ListUsersHandler)).Methods("GET")
-	r.HandleFunc("/v1/users/{id}", contextAndErrors(api.GetUserHandler)).Methods("GET")
-	r.HandleFunc("/v1/users/{id}", contextAndErrors(api.UpdateUserHandler)).Methods("PUT")
+	r.HandleFunc("/v1/tokens/self", contextAndErrors(api.SignOutHandler)).Methods(http.MethodDelete)
+	r.HandleFunc("/v1/tokens/self", contextAndErrors(api.RefreshHandler)).Methods(http.MethodPut)
+	r.HandleFunc("/v1/users", contextAndErrors(api.CreateUserHandler)).Methods(http.MethodPost)
+	r.HandleFunc("/v1/users", contextAndErrors(api.ListUsersHandler)).Methods(http.MethodGet)
+	r.HandleFunc("/v1/users/{id}", contextAndErrors(api.GetUserHandler)).Methods(http.MethodGet)
+	r.HandleFunc("/v1/users/{id}", contextAndErrors(api.UpdateUserHandler)).Methods(http.MethodPut)
 	// self used in paths rather than identifier as the identifier is a Cognito Session string in change password requests
 	// the user id is not yet available from the previous responses
-	r.HandleFunc("/v1/users/self/password", contextAndErrors(api.ChangePasswordHandler)).Methods("PUT")
-	r.HandleFunc("/v1/password-reset", contextAndErrors(api.PasswordResetHandler)).Methods("POST")
+	r.HandleFunc("/v1/users/self/password", contextAndErrors(api.ChangePasswordHandler)).Methods(http.MethodPut)
+	r.HandleFunc("/v1/password-reset", contextAndErrors(api.PasswordResetHandler)).Methods(http.MethodPost)
 	return api, nil
 }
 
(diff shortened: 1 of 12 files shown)


api/users.go
```diff
@@ -112,6 +112,60 @@ func (api *API) GetUserHandler(ctx context.Context, w http.ResponseWriter, req *
...
+		return nil, models.NewErrorResponse(validationErrs, http.StatusBadRequest, nil)
+	}
+
+	userRequest := user.BuildUpdateUserRequest(api.UserPoolId)
+
+	_, err = api.CognitoClient.AdminUpdateUserAttributes(userRequest)
+	if err != nil {
+		responseErr := models.NewCognitoError(ctx, err, "AdminUpdateUserAttributes request from update user endpoint")
+		if responseErr.Code == models.UserNotFoundError {
+			return nil, models.NewErrorResponse([]error{responseErr}, http.StatusNotFound, nil)
```
daiLlew: May as well create a var for this: `[]error{responseErr}` seeing as you always need it in that format just make its easier to read.

api/users.go
```diff
@@ -112,6 +112,60 @@ func (api *API) GetUserHandler(ctx context.Context, w http.ResponseWriter, req *
...
+		} else if responseErr.Code == models.InvalidFieldError {
+			return nil, models.NewErrorResponse([]error{responseErr}, http.StatusBadRequest, nil)
+		} else {
+			return nil, models.NewErrorResponse([]error{responseErr}, http.StatusInternalServerError, nil)
+		}
```
daiLlew: I don't think the `else` is required here, same is achieved if you just fall through to this return statement.

api/users.go
```diff
@@ -112,6 +112,60 @@ func (api *API) GetUserHandler(ctx context.Context, w http.ResponseWriter, req *
...
+	}
+
+	userDetailsRequest := user.BuildAdminGetUserRequest(api.UserPoolId)
+	userDetailsResponse, err := api.CognitoClient.AdminGetUser(userDetailsRequest)
+	if err != nil {
+		responseErr := models.NewCognitoError(ctx, err, "AdminGetUser request from update user endpoint")
+		return nil, models.NewErrorResponse([]error{responseErr}, http.StatusInternalServerError, nil)
```
daiLlew, line 156: This is probably beyond the scope of this PR but I am thinking adding a new func to make this syntax a bit nice might be a good idea:
Atm we have:
```
func NewErrorResponse(errors []error, statusCode int, headers map[string]string) *ErrorResponse {}
```
But that requires you to do the following all the time when you only have 1 error.
```
[]error{responseErr}
```
Its driving my OCD wild 😂  We could add another "overloaded func":
```
func NewErrorResponse(statusCode int, headers map[string]string, errors... error) *ErrorResponse {}
```
Which would be a bit nicer to use. Again totally beyond the scope of this PR.

api/users_test.go
```diff
@@ -380,6 +380,156 @@ func TestGetUserHandler(t *testing.T) {
...
+				http.StatusInternalServerError,
+			},
+		}
+
+		for _, tt := range adminCreateUsersTests {
+			m.AdminUpdateUserAttributesFunc = tt.updateUserFunction
+			m.AdminGetUserFunc = tt.getUserFunction
+
+			postBody := map[string]interface{}{"forename": tt.userForename, "lastname": lastname, "status": status, "role_type": roleType}
+			body, _ := json.Marshal(postBody)
```
daiLlew: Should check this error, can do a `So(err, ShouldBeNil)` which will fail the test.

api/users_test.go
```diff
@@ -380,6 +380,156 @@ func TestGetUserHandler(t *testing.T) {
...
+			r := httptest.NewRequest(http.MethodGet, userEndPoint, bytes.NewReader(body))
+
+			successResponse, errorResponse := api.UpdateUserHandler(ctx, w, r)
+
+			// Check whether testing a success or error case
+			if tt.httpResponse > 399 {
```
daiLlew: this is relying on the tests returning the right responses for each test case. Is there a way we can add the "expectation" to the test case struct? That way if the test returns an incorrect response the expectation defined in the test case will ensure the test fails appropriately.

mwnicks: This was actually relying on response code a set in the test case struct to determine which set of assertions to user for the test case.
But I agree having the assertions as part of the test case struct makes it cleaner and easier to follow.

api/users.go
mwnicks: This has actually been bugging me too.
The suggested approach is definitely cleaner.
I'll open a separate PR to refactor this across the service

api/api.go
```diff
@@ -69,6 +69,7 @@ func Setup(ctx context.Context, r *mux.Router, cognitoClient cognito.Client, use
 	r.HandleFunc("/v1/users", contextAndErrors(api.CreateUserHandler)).Methods("POST")
 	r.HandleFunc("/v1/users", contextAndErrors(api.ListUsersHandler)).Methods("GET")
 	r.HandleFunc("/v1/users/{id}", contextAndErrors(api.GetUserHandler)).Methods("GET")
+	r.HandleFunc("/v1/users/{id}", contextAndErrors(api.UpdateUserHandler)).Methods("PUT")
```
mike-kelly-methods: I'm thinking Matt, it'd probably be good to get some consistency by replacing the inlining of the http verb with `net/http` constants as is the case in other areas of the codebase:

`import net/http`
...
...
...
`r.HandleFunc("/v1/users/{id}", contextAndErrors(api.UpdateUserHandler)).Methods(http.MethodPut)`

daiLlew (approved):

api/api_test.go
```diff
@@ -39,6 +39,7 @@ func TestSetup(t *testing.T) {
 			So(hasRoute(api.Router, "/v1/users", "POST"), ShouldBeTrue)
 			So(hasRoute(api.Router, "/v1/users", "GET"), ShouldBeTrue)
 			So(hasRoute(api.Router, "/v1/users/{id}", "GET"), ShouldBeTrue)
+			So(hasRoute(api.Router, "/v1/users/{id}", "PUT"), ShouldBeTrue)
```
mike-kelly-methods, line 42: Not a biggy - but we should use the `net/http` constants here too.

mike-kelly-methods (approved):
👍

PR #52, opened 2021-07-05
Feature/role group initialisation

mwnicks:
### What

Add check for role groups, adding them if missing, to api setup.

### How to review

Ensure the service can be started and run correctly.

### Who can review

Anyone but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -132,3 +132,27 @@ func handleBodyUnmarshalError(ctx context.Context, err error) *models.ErrorRespo
 		nil,
 	)
 }
+
+func initialiseRoleGroups(ctx context.Context, cognitoClient cognito.Client, userPoolId string) error {
+	adminGroup := models.NewAdminRoleGroup()
+	adminGroupCreateInput := adminGroup.BuildCreateGroupRequest(userPoolId)
+	_, err := cognitoClient.CreateGroup(adminGroupCreateInput)
+	if err != nil {
+		cognitoErr := models.NewCognitoError(ctx, err, "CreateGroup request for admin group from API start up")
+		if cognitoErr.Code != models.GroupExistsError {
+			return cognitoErr
+		}
+	}
+
+	publisherGroup := models.NewPublisherRoleGroup()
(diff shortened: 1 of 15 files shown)


api/api.go
```diff
@@ -132,3 +136,27 @@ func handleBodyUnmarshalError(ctx context.Context, err error) *models.ErrorRespo
...
+
+func initialiseRoleGroups(ctx context.Context, cognitoClient cognito.Client, userPoolId string) error {
+	adminGroup := models.NewAdminRoleGroup()
+	adminGroupCreateInput := adminGroup.BuildCreateGroupRequest(userPoolId)
+	_, err := cognitoClient.CreateGroup(adminGroupCreateInput)
+	if err != nil {
+		cognitoErr := models.NewCognitoError(ctx, err, "CreateGroup request for admin group from API start up")
+		if cognitoErr.Code != models.GroupExistsError {
+			return cognitoErr
+		}
```
daiLlew, line 148: Just making sure I follow this logic: If there is an error we only return it if its a Group exists error?

models/groups.go
```diff
@@ -0,0 +1,46 @@
+package models
+
+import "github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
+
+const (
+	AdminRoleGroup     = "role-admin"
+	PublisherRoleGroup = "role-publisher"
+)
+
+type Group struct {
```
daiLlew, line 11: Godoc on public types.

api/api.go
mwnicks: No we return any errors that aren't `GroupExistsError`.
`GroupExistsError`s are swallowed.

The purpose of this function is to make sure the role groups are setup in Cognito, creating them if they aren't there.
If Cognito returns a success or a `GroupExistsError` then the group is in place and the service is ok to run. Any other error needs to be returned and stop the service starting.

models/groups.go
mwnicks: added Godocs for the type and functions

daiLlew (approved):

PR #54, opened 2021-07-09
change constructor to accept one or more errors to make calls cleaner

mwnicks:
### What

Refactor the ErrorResponse constructor to accept one or more errors, rather than a list, to make function calls cleaner when there is only one error.

### How to review

Ensure all unit and component tests are still passing.

### Who can review

Anyone but me


DavidSubiros (dismissed):

mike-kelly-methods (approved):
LGTM 👍

PR #57, opened 2021-07-12
Feature/update user state

mwnicks:
### What

Update the update user endpoint to enable/disable users as well.

### How to review

Send a request to PUT '/v1/users/{user_id}' with user_id = ca83ca4c-0495-4254-a131-57ada4b1fc70 and request body:
- forename = a new forename for the user
- lastname = a new surname for the user
- active = false

You should get a 200 response with the response body like:
{
    "forename": <value set in request>,
    "lastname": <value set in request>,
    "email": "matt.nicks@ons.gov.uk",
    "groups": [],
    "status": "CONFIRMED",
    "active": false,
    "id": "ca83ca4c-0495-4254-a131-57ada4b1fc70",
    "status_notes": ""
}

### Who can review

Anyone but me

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -3,10 +3,10 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/gorilla/mux"
 	"io/ioutil"
 	"net/http"
-
-	"github.com/gorilla/mux"
 
 	"github.com/ONSdigital/dp-identity-api/models"
 	"github.com/google/uuid"
@@ -59,7 +59,7 @@ func (api *API) CreateUserHandler(ctx context.Context, w http.ResponseWriter, re
 			return nil, models.NewErrorResponse(http.StatusBadRequest, nil, responseErr)
 		}
 	}
-
+	resultUser.User.Enabled = aws.Bool(true)
 	createdUser := models.UserParams{}.MapCognitoDetails(resultUser.User)
 	jsonResponse, responseErr := createdUser.BuildSuccessfulJsonResponse(ctx)
 	if responseErr != nil {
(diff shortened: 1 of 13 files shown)


cognito/mock/user_mocker.go
```diff
@@ -47,5 +49,19 @@ func (m *CognitoIdentityProviderClientStub) GenerateUser(id, email, password, gi
...
+}
+
+func (m *CognitoIdentityProviderClientStub) SetUserActiveState(username, active string) {
+	for _, user := range m.Users {
+		if user.ID == username {
+			if active == "true" {
+				user.Active = true
+			} else {
+				user.Active = true
+			}
```
daiLlew: Should this be `else user.Active = false`?  You could simplify this to:
```
user.Active = "true" == active
```

api/users.go
```diff
@@ -136,17 +136,23 @@ func (api *API) UpdateUserHandler(ctx context.Context, w http.ResponseWriter, re
 		return nil, models.NewErrorResponse(http.StatusBadRequest, nil, validationErrs...)
 	}
 
-	userRequest := user.BuildUpdateUserRequest(api.UserPoolId)
+	if user.Active {
+		userEnableRequest := user.BuildEnableUserRequest(api.UserPoolId)
```
sandypadmanabhan, line 140: can UserPoolId be null or empty?

mwnicks: No I don't think it can be.
When the API is setup the userPoolId is set (along with some other required config) and if the value isn't there is throws and error and shuts down.
https://github.com/ONSdigital/dp-identity-api/blob/develop/api/api.go#L52

sandypadmanabhan (approved):
LGTM

PR #61, opened 2021-07-22
Feature/remove user from group

mwnicks:
### What

Add a remove user from group endpoint.

### How to review

Ensure all the unit and feature test are passing
Send a POST request to /v1/groups/role-publisher/members/ca83ca4c-0495-4254-a131-57ada4b1fc70, with body:

You should receive a 200 response with body that looks like:
{
"name": "role-publisher",
"description": "The publishers",
"precedence": 1,
"created": "2021-07-08T16:59:24.---Z"
"members": []
}

### Who can review

Anyone but me

### DIFF
--- a/api/groups.go
+++ b/api/groups.go
@@ -112,5 +112,5 @@ func (api *API) RemoveUserFromGroupHandler(ctx context.Context, w http.ResponseW
 		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, responseErr)
 	}
 
-	return models.NewSuccessResponse(jsonResponse, http.StatusNoContent, nil), nil
+	return models.NewSuccessResponse(jsonResponse, http.StatusOK, nil), nil
 }
--- a/api/groups.go
+++ b/api/groups.go
@@ -26,7 +26,7 @@ func (api *API) AddUserToGroupHandler(ctx context.Context, w http.ResponseWriter
 	}
 	userId := bodyJson["user_id"]
 
-	validationErrs := group.ValidateAddUser(ctx, userId)
+	validationErrs := group.ValidateAddRemoveUser(ctx, userId)
 	if len(validationErrs) != 0 {
 		return nil, models.NewErrorResponse(http.StatusBadRequest, nil, validationErrs...)
 	}
--- a/api/groups.go
+++ b/api/groups.go
@@ -66,3 +66,51 @@ func (api *API) AddUserToGroupHandler(ctx context.Context, w http.ResponseWriter
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusOK, nil), nil
 }
+
+//RemoveUserFromGroupHandler adds a user to the specified group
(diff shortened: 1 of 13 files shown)


api/groups.go
```diff
@@ -66,3 +66,51 @@ func (api *API) AddUserToGroupHandler(ctx context.Context, w http.ResponseWriter
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusOK, nil), nil
 }
+
+//RemoveUserFromGroupHandler adds a user to the specified group
+func (api *API) RemoveUserFromGroupHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	vars := mux.Vars(req)
+	group := models.Group{Name: vars["id"]}
```
mike-kelly-methods, line 73: Are we always certain that the`.../{id}/...`  param passed to the handler represents a valid group?

api/groups_test.go
```diff
@@ -257,3 +258,243 @@ func TestAddUserToGroupHandler(t *testing.T) {
...
+						Group: getGroupData,
+					}
+					return group, nil
+				},
+				func(input *cognitoidentityprovider.ListUsersInGroupInput) (*cognitoidentityprovider.ListUsersInGroupOutput, error) {
+					return &cognitoidentityprovider.ListUsersInGroupOutput{}, nil
+				},
+				func(successResponse *models.SuccessResponse, errorResponse *models.ErrorResponse) {
+					So(successResponse, ShouldNotBeNil)
+					So(successResponse.Status, ShouldEqual, http.StatusOK)
```
mike-kelly-methods, line 301: This should be `http.StatusAccepted` for a `202` (or maybe the comment at the top of this test item is incorrect?)

api/groups_test.go
```diff
@@ -257,3 +258,243 @@ func TestAddUserToGroupHandler(t *testing.T) {
...
+					So(errorResponse, ShouldBeNil)
+				},
+			},
+			// Cognito 404 response - user not found
```
mike-kelly-methods, line 305: Looking at the response, and the swagger doc, I think this should be `400`

mwnicks: This comment is describing the response from Cognito not the response from the API.
Would probably be less confusing to leave the Cognito status code out of the comment

api/groups_test.go
mwnicks: It was a 202 response, I changed it and must have missed updating the comment.
Will update now.

api/groups_test.go
mwnicks: I think I've misunderstood your comment here.

api/groups_test.go
mike-kelly-methods: 👍

api/groups_test.go
mike-kelly-methods: My bad - its the 404 from Cognito

mike-kelly-methods (approved):
👍

PR #62, opened 2021-07-23
Feature/configurable email validation

mwnicks:
### What

Made the allowed domains used to verify email address of new users a config attribute, so it can be set differently in different environments

### How to review

Ensure all unit and feature tests are still passing and the services runs.
Send a POST request to /v1/users with body:
{
    "forename": "J",
    "lastname": "Doe",
    "email":  (any non ons.gov.uk or ext.ons.gov.uk email address)
}
You should receive a 400 response with an InvalidEmail error.

Locally set the env variable ALLOWED_EMAIL_DOMAINS as the domain of your non ons email address.
Restart the service and resend the request from above.

You should receive a 201 response this time.

### Who can review

Anyone but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -29,6 +29,7 @@ type API struct {
 	ClientId       string
 	ClientSecret   string
 	ClientAuthFlow string
+	AllowedDomains []string
 }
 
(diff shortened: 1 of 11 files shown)


HeJones (approved):
It looks good to me and the behaviour seen in Postman was as described

PR #64, opened 2021-07-27
Add users global signout handler for processing bulk signout event.

mike-kelly-methods:
### What

As per title - handle bulk `panic` signout request.

### How to review

Review code and ensure unit tests are passing.

### Who can review

Anyone but me.

### DIFF
--- a/api/tokens.go
+++ b/api/tokens.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/ONSdigital/dp-identity-api/models"
-	"github.com/ONSdigital/log.go/log"
 	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
 )
 
@@ -158,7 +157,8 @@ func (api *API) SignOutAllUsersHandler(ctx context.Context, w http.ResponseWrite
 	listUserInput := usersList.BuildListUserRequest("status=\"Enabled\"", "username", int64(0), &api.UserPoolId)
 	listUserResp, err := api.CognitoClient.ListUsers(listUserInput)
 	if err != nil {
-		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, err)
+		responseError := models.NewCognitoError(ctx, err, "Cognito ListUsers request from signout all users from group endpoint")
+		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, responseError)
 	}
 	usersList.MapCognitoUsers(listUserResp)
 	globalSignOut := &models.GlobalSignOut{
@@ -171,52 +171,82 @@ func (api *API) SignOutAllUsersHandler(ctx context.Context, w http.ResponseWrite
 		RetryAllowed: true,
 	}
 	// run api.SignOutUsersWorker concurrently
-	go api.SignOutUsersWorker(ctx, globalSignOut, &usersList.Users)
+	go api.SignOutUsersWorker(req.Context(), globalSignOut, &usersList.Users)
 	return models.NewSuccessResponse(nil, http.StatusAccepted, nil), nil
 }
 
 // SignOutUsersWorker - signs out users globally by invalidating user's refresh token
 func (api *API) SignOutUsersWorker(ctx context.Context, g *models.GlobalSignOut, usersList *[]models.UserParams) {
 	userSignOutRequestData := g.BuildSignOutUserRequest(usersList, &api.UserPoolId)
-	for i := range userSignOutRequestData {
+	
+	for _, userSignoutRequest := range userSignOutRequestData {
+
 		for _, backoff := range g.BackoffSchedule {
-			_, err := api.generateGlobalSignOutRequest(userSignOutRequestData[i])
-			if err != nil {
+			_, err := api.generateGlobalSignOutRequest(userSignoutRequest)
+
+			// no errors returned - add username to results channel and break to next user in list
+			if err == nil {
+
+				// the results channel is not being processed by caller currently - here for possible future extensibility
+				g.ResultsChannel <- *userSignoutRequest.Username
+				g.RetryAllowed = true
+				break
+
+			} else {
+
+				// error returned - process it
 				responseErr := models.NewCognitoError(ctx, err, "Cognito AdminUserGlobalSignOut request for sign out")
-				if responseErr.Code != models.TooManyRequestsError { // 1. Process all errors other than TooManyRequestsException (429)
-					if g.RetryAllowed { // 2. Attempt one more request to AdminUserGlobalSignOut if GlobalSignOut.RetryAllowed true, else break to next user
+
+				// is error code != models.TooManyRequestsError? If so, proceed...
+				if responseErr.Code != models.TooManyRequestsError {
+
+					// if g.RetryAllowed is true, allowed to request api again
+					if g.RetryAllowed {
+
+						// attempt one more request to api
 						g.RetryAllowed = false // 3. Set GlobalSignOut.RetryAllowed to false and request AdminUserGlobalSignOut
-						log.Event(ctx, "Error Cognito AdminUserGlobalSignOut:", log.ERROR, log.Error(err))
-						_, retryErr := api.generateGlobalSignOutRequest(userSignOutRequestData[i])
-						if retryErr != nil { // 4. If error response from request received, process it
+						_, retryErr := api.generateGlobalSignOutRequest(userSignoutRequest)
+
+						if retryErr != nil {
+
+							// if error response from request received again, process it
 							retryResponseErr := models.NewCognitoError(ctx, err, "Cognito AdminUserGlobalSignOut request for sign out")
-							if retryResponseErr.Code != models.TooManyRequestsError { // 4.1 If error is not a TooManyRequestsException (429), reset GlobalSignOut.RetryAllowed and break to next user
+
+							 // if error code != models.TooManyRequestsError break to next user
+							if retryResponseErr.Code != models.TooManyRequestsError {
 								g.RetryAllowed = true
 								break
 							}
-						} else { // 5. If request successful, reset GlobalSignOut.RetryAllowed and set original response error to nil - username will be added to GlobalSignOut.ResultsChannel
+
+						} else {
+
+							// no error on retry, add user to results channel and break to next user
+							// the results channel is not being processed by caller currently - here for possible future extensibility
+							g.ResultsChannel <- *userSignoutRequest.Username
 							g.RetryAllowed = true
-							err = nil
+							break
+
 						}
-					} else { // 6. If GlobalSignOut.RetryAllowed is already false, know that a second request already made, so reset GlobalSignOut.RetryAllowed and break to next user 
+
+					} else { 
+
+						// if GlobalSignOut.RetryAllowed already false break to next user
 						g.RetryAllowed = true
 						break
+
 					}
 				}
 			}
-			if err == nil {
-				g.ResultsChannel <- *userSignOutRequestData[i].Username
-				break
-			}
-			log.Event(ctx, "Error Cognito AdminUserGlobalSignOut:", log.ERROR, log.Error(err))
+			
+			// backoff for predetermined length of time before requesting again
 			time.Sleep(backoff)
 		}
+		
 	}
 	close(g.ResultsChannel)
 }
 
 // generateGlobalSignOutRequest - local routine to generete the global signout request per user
 func (api *API) generateGlobalSignOutRequest(user *cognitoidentityprovider.AdminUserGlobalSignOutInput) (*cognitoidentityprovider.AdminUserGlobalSignOutOutput, error) {
-	response, error := api.CognitoClient.AdminUserGlobalSignOut(user)
-	return response, error
+	return api.CognitoClient.AdminUserGlobalSignOut(user)
 }
--- a/api/tokens.go
+++ b/api/tokens.go
@@ -179,7 +179,6 @@ func (api *API) SignOutAllUsersHandler(ctx context.Context, w http.ResponseWrite
 func (api *API) ListUsersWorker(ctx context.Context, userFilterString *string, backoffSchedule []time.Duration) (*[]models.UserParams, *models.ErrorResponse) {
 	var (
 		awsErr error
-		paginationToken *string
 		usersList models.UsersList
 		listUsersResp, result *cognitoidentityprovider.ListUsersOutput
 		listUserInput = usersList.BuildListUserRequest(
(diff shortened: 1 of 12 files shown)


api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusCreated, headers), nil
 }
+
+//SignOutAllUsersHandler bulk refresh token invalidation for panic sign out handling
+func (api *API) SignOutAllUsersHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
```
dandwal, line 155: I think context is a property of the http.request so we are technically passing in context as an argument twice, we could remove the first context argument and then when needing to use the context take it from the request instead `req.context`

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+	userSignOutRequestData := g.BuildSignOutUserRequest(usersList, &api.UserPoolId)
+	for i := range userSignOutRequestData {
+		for _, backoff := range g.BackoffSchedule {
+			_, err := api.generateGlobalSignOutRequest(userSignOutRequestData[i])
+			if err != nil {
+				responseErr := models.NewCognitoError(ctx, err, "Cognito AdminUserGlobalSignOut request for sign out")
+				if responseErr.Code != models.TooManyRequestsError { // 1. Process all errors other than TooManyRequestsException (429)
+					if g.RetryAllowed { // 2. Attempt one more request to AdminUserGlobalSignOut if GlobalSignOut.RetryAllowed true, else break to next user
+						g.RetryAllowed = false // 3. Set GlobalSignOut.RetryAllowed to false and request AdminUserGlobalSignOut
+						log.Event(ctx, "Error Cognito AdminUserGlobalSignOut:", log.ERROR, log.Error(err))
```
ChrysmOre: This is all pretty hard to follow at first glance. Consider using some white space to break up the code into logical paragraphs.

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+							g.RetryAllowed = true
+							err = nil
+						}
+					} else { // 6. If GlobalSignOut.RetryAllowed is already false, know that a second request already made, so reset GlobalSignOut.RetryAllowed and break to next user 
+						g.RetryAllowed = true
+						break
+					}
+				}
+			}
+			if err == nil {
```
ChrysmOre: Instead of an `if err != nil` clause followed by a `if err == nil` clause, it would probably be better to use an `else` clause

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+			log.Event(ctx, "Error Cognito AdminUserGlobalSignOut:", log.ERROR, log.Error(err))
+			time.Sleep(backoff)
+		}
+	}
+	close(g.ResultsChannel)
+}
+
+// generateGlobalSignOutRequest - local routine to generete the global signout request per user
+func (api *API) generateGlobalSignOutRequest(user *cognitoidentityprovider.AdminUserGlobalSignOutInput) (*cognitoidentityprovider.AdminUserGlobalSignOutOutput, error) {
+	response, error := api.CognitoClient.AdminUserGlobalSignOut(user)
```
ChrysmOre: Looks like you could just `api.CognitoClient.AdminUserGlobalSignOut(user)` here

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+		ResultsChannel: make(chan string, len(usersList.Users)),
+		BackoffSchedule: []time.Duration{
+			1 * time.Second,
+			3 * time.Second,
+			10 * time.Second,
+		},
+		RetryAllowed: true,
+	}
+	// run api.SignOutUsersWorker concurrently
+	go api.SignOutUsersWorker(ctx, globalSignOut, &usersList.Users)
```
ChrysmOre: Firing off a goroutine without knowing how or when it will end seems like potentially undefined behaviour to me.. was there a reason this was done concurrently?

api/tokens_test.go
```diff
@@ -363,3 +364,199 @@ func TestAPI_RefreshHandler(t *testing.T) {
...
+
+			successResponse, errorResponse := api.SignOutAllUsersHandler(ctx, w, r)
+			So(successResponse, ShouldBeNil)
+			So(errorResponse.Status, ShouldEqual, tt.httpResponse)
+		}
+	})
+}
+
+func TestSignOutAllUsersGoRoutine(t *testing.T) {
+
```
ChrysmOre: Seems to be an inconsistent blank line here

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+							err = nil
+						}
+					} else { // 6. If GlobalSignOut.RetryAllowed is already false, know that a second request already made, so reset GlobalSignOut.RetryAllowed and break to next user 
+						g.RetryAllowed = true
+						break
+					}
+				}
+			}
+			if err == nil {
+				g.ResultsChannel <- *userSignOutRequestData[i].Username
```
ChrysmOre: What happens after this is sent to this channel? I can't see where it's being received

ChrysmOre (changes requested):
Would strongly recommend using white space to split the code up into readable chunks throughout.
I have some questions about what's happening in the code.
Else clause should be used instead of 2 if statements in a row.

api/tokens.go
mike-kelly-methods: Yes - thought it would be cleaner to have a private package function call the api endpoint  and call that, rather than have two separate api calls in the same function.

api/tokens.go
mike-kelly-methods: The decision was taken to run this go routine in non-blocking asynchronous mode - and let it handle the calls to to the api without the caller having to handle any returns/errors.

api/tokens_test.go
mike-kelly-methods: 👍

api/tokens.go
mike-kelly-methods: As the function is being run asynchronously, we're not overly concerned about the handling data at the moment, but we could be at a point in the future should this be required.

Its also being used in unit testing.

api/tokens.go
mike-kelly-methods: Yes, of course. 👍

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
 	return models.NewSuccessResponse(jsonResponse, http.StatusCreated, headers), nil
 }
+
+//SignOutAllUsersHandler bulk refresh token invalidation for panic sign out handling
+func (api *API) SignOutAllUsersHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	usersList := models.UsersList{}
+	listUserInput := usersList.BuildListUserRequest("status=\"Enabled\"", "username", int64(0), &api.UserPoolId)
+	listUserResp, err := api.CognitoClient.ListUsers(listUserInput)
+	if err != nil {
+		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, err)
```
daiLlew: Add a log error event before returning here - it gives us some context what the app was trying to do before it failed.

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+		},
+		RetryAllowed: true,
+	}
+	// run api.SignOutUsersWorker concurrently
+	go api.SignOutUsersWorker(ctx, globalSignOut, &usersList.Users)
+	return models.NewSuccessResponse(nil, http.StatusAccepted, nil), nil
+}
+
+// SignOutUsersWorker - signs out users globally by invalidating user's refresh token
+func (api *API) SignOutUsersWorker(ctx context.Context, g *models.GlobalSignOut, usersList *[]models.UserParams) {
```
daiLlew, line 236: Observations from me.
- We are calling this method from a Go routine so the API request is async. But we are iterating through the users within that go routine in series. I'm unaware of the requirements discussions you've already had so just wondering if this was a requirement or decision made intentionally?

   If not (at this risk of making this even more complex) I am wondering if we can/should make this process concurrent as well? i.e. process the list of users concurrently within this go routine?

- This is a complex block of code and there is a lot going on here. I think we should consider breaking this down a bit into smaller methods to make it a bit easier to digest.

I appreciate this is a complex problem with very restrictive requirements Mike and you've invested time and effort into working out how to solve it. Im also unfamiliar with the nuances of it so forgive me for asking any dumb questions.

api/tokens.go
ChrysmOre: Ok. It might be worth adding a comment here that the results channel isn't being used in the main code as of yet, as it doesn't make sense to send something with no plan to ever receive it.

api/tokens.go
ChrysmOre: Yeah I wasn't suggesting any change in structure just that those 2 lines could be written as one - Ah I've just noticed my comment was missing the `return`. What I meant to say was:

Looks like you could just ` return api.CognitoClient.AdminUserGlobalSignOut(user)` here

mike-kelly-methods: 👍

api/tokens.go
mike-kelly-methods: 👍

api/tokens.go
mike-kelly-methods: Thanks Dave. Yes, good point. I discussed with James. The AWS Cognito `AdminUsersGlobalSignOut` API restricts requests to `25 requests/second` - see:

https://docs.aws.amazon.com/cognito/latest/developerguide/limits.html#:~:text=and%20user%20attributes%2C-,25,-UserResourceRead.

We felt that distributing the workload to `n` number of workers would be negated by this request limit. Thats why I've implemented the exponential backoff feature as this single concurrent worker will hit that and will have to backoff to ensure that it adheres to this limit too to avoid DDOS and possible permanent blacklisting.

api/tokens.go
mwnicks: We should be mapping the Cognito error to the internal error 'codes' before returning.
`models.NewCognitoError(err)`
this will handle the error logging too.

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusCreated, headers), nil
 }
+
+//SignOutAllUsersHandler bulk refresh token invalidation for panic sign out handling
+func (api *API) SignOutAllUsersHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	usersList := models.UsersList{}
+	listUserInput := usersList.BuildListUserRequest("status=\"Enabled\"", "username", int64(0), &api.UserPoolId)
+	listUserResp, err := api.CognitoClient.ListUsers(listUserInput)
```
mwnicks: Whilst working on my current ticket I've come across a hard limit  on the ListUsers request that isn't document in the AWS Cognito SDK docs. The most it will ever return is 60 users per request.
Found [here](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListUsers.html) in the API documentation.
I'm going to drop a note in the 404 channel too as this affects a number of endpoints and the backup restores.

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+	}
+	// run api.SignOutUsersWorker concurrently
+	go api.SignOutUsersWorker(ctx, globalSignOut, &usersList.Users)
+	return models.NewSuccessResponse(nil, http.StatusAccepted, nil), nil
+}
+
+// SignOutUsersWorker - signs out users globally by invalidating user's refresh token
+func (api *API) SignOutUsersWorker(ctx context.Context, g *models.GlobalSignOut, usersList *[]models.UserParams) {
+	userSignOutRequestData := g.BuildSignOutUserRequest(usersList, &api.UserPoolId)
+	for i := range userSignOutRequestData {
```
mwnicks: Could this not be 
`for _, userSignoutRequest := range userSignoutRequestData`
rather than using the index in the loop?

api/tokens.go
```diff
@@ -148,3 +151,72 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+						g.RetryAllowed = true
+						break
+					}
+				}
+			}
+			if err == nil {
+				g.ResultsChannel <- *userSignOutRequestData[i].Username
+				break
+			}
+			log.Event(ctx, "Error Cognito AdminUserGlobalSignOut:", log.ERROR, log.Error(err))
```
mwnicks: This error is already being logged on line 185 when `models.NewCognitoError()` is called

api/tokens.go
mwnicks: This error is already being logged on line 185 when models.NewCognitoError() is called

api/tokens.go
mwnicks: I think this is an `if` rather than an `else` so if the retry succeeds `err` is reset to `nil` on line 199 and this `if` can then catch both.
But i think it might be sacrificing readability to not repeat 2 lines.

Reordering the if check and repeating the lines would remove a layer of nesting and might improve readability

```suggestion
if err == nil {
	g.ResultsChannel <- *userSignOutRequestData[i].Username
	break
}
responseErr := models.NewCognitoError(ctx, err, "Cognito AdminUserGlobalSignOut request for sign out")
if responseErr.Code != models.TooManyRequestsError {
	if g.RetryAllowed { 
		g.RetryAllowed = false
		_, retryErr := api.generateGlobalSignOutRequest(userSignOutRequestData[i])
		if retryErr != nil { 
			retryResponseErr := models.NewCognitoError(ctx, err, "Cognito AdminUserGlobalSignOut request for sign out")
			if retryResponseErr.Code != models.TooManyRequestsError { 
				g.RetryAllowed = true
				break
			}
		} else { 
			g.RetryAllowed = true
			g.ResultsChannel <- *userSignOutRequestData[i].Username
			break
		}
	} else { 
		g.RetryAllowed = true
		break
	}
}
time.Sleep(backoff)
```

mwnicks (changes requested):

api/tokens.go
```diff
@@ -148,3 +150,171 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+		listUserInput = usersList.BuildListUserRequest(
+			*userFilterString,
+			"",
+			int64(0), 
+			nil,
+			&api.UserPoolId,
+		)
+		usersListError *models.ErrorResponse
+	)
+	listUsersResp, awsErr = api.generateListUsersRequest(listUserInput)
```
mwnicks, line 193: I think this could be simplified to something like this
```suggestion
	loadingInProgress := true
	for loadingInProgress {
		for _, backoff := range backoffSchedule {
			result, awsErr = api.generateListUsersRequest(listUserInput)
			if awsErr == nil {
				listUsersResp.Users = append(listUsersResp.Users, result.Users...)
				if result.PaginationToken != nil {
					listUserInput.PaginationToken = result.PaginationToken
				} else {
					loadingInProgress = false
				}
			} else {
				err := models.NewCognitoError(ctx, awsErr, "Cognito ListUsers request from signout all users from group endpoint")
				if err.Code != models.TooManyRequestsError {
					return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, err)
				}
				time.Sleep(backoff)
			}
		}
	}
	usersList.MapCognitoUsers(listUsersResp)
	return &usersList.Users, nil
```

mwnicks (dismissed):

mwnicks (dismissed):

api/tokens.go
```diff
@@ -148,3 +150,165 @@ func (api *API) RefreshHandler(ctx context.Context, w http.ResponseWriter, req *
...
+				break
+
+			} else {
+
+				// error returned - process it
+				responseErr := models.NewCognitoError(ctx, err, "Cognito AdminUserGlobalSignOut request for sign out")
+
+				// is error code != models.TooManyRequestsError? If so, proceed...
+				if responseErr.Code != models.TooManyRequestsError {
+
```
ChrysmOre: I see you took my comment about whitespace to heart :) There's no need to add blank lines at the beginning or end of a block, but otherwise looks better

ChrysmOre (approved):

PR #67, opened 2021-08-09
Feature/list users group endpoint

HeJones:
### What

Add endpoint 
GET /v1/groups/{id}/members
to list the users belonging to a group

### How to review

Ensure unit tests and component tests pass

Run `make debug`:
when you make a GET request to the following endpoint: http://localhost:25600/v1/groups/role-admins/members you should see 1 users
when you make a GET request to the following endpoint: http://localhost:25600/v1/groups/test-group-0/members you should see 178 users (there is a count variable at the bottom of the list)
### Who can review

Anyone

### DIFF
--- a/api/groups.go
+++ b/api/groups.go
@@ -113,10 +113,9 @@ func (api *API) getUsersInAGroup(listOfUsers []*cognitoidentityprovider.UserType
 		}
 
 		listOfUsers = append(listOfUsers, groupMembersResponse.Users...)
+		nextToken = ""
 		if groupMembersResponse.NextToken != nil {
 			nextToken = *groupMembersResponse.NextToken
-		} else {
-			nextToken = ""
 		}
 	}
 	return listOfUsers, nil
--- a/api/groups.go
+++ b/api/groups.go
@@ -75,7 +75,9 @@ func (api *API) ListUsersInGroupHandler(ctx context.Context, w http.ResponseWrit
 	vars := mux.Vars(req)
 	group := models.Group{Name: vars["id"]}
 
-	listUsers, err := api.getUsersInAGroup(nil, group)
+	listOfUsersInput := []*cognitoidentityprovider.UserType{}
+
+	listUsers, err := api.getUsersInAGroup(listOfUsersInput, group)
 	if err != nil {
 		cognitoErr := models.NewCognitoError(ctx, err, "Cognito ListUsersInGroup request from list users in group endpoint")
 		if cognitoErr.Code == models.NotFoundError {
--- a/api/groups.go
+++ b/api/groups.go
@@ -75,10 +75,7 @@ func (api *API) ListUsersInGroupHandler(ctx context.Context, w http.ResponseWrit
 	vars := mux.Vars(req)
 	group := models.Group{Name: vars["id"]}
 
-	groupMembersRequest := group.BuildListUsersInGroupRequest(api.UserPoolId)
-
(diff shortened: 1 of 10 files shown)


api/groups.go
```diff
@@ -67,6 +69,59 @@ func (api *API) AddUserToGroupHandler(ctx context.Context, w http.ResponseWriter
...
+		firstTimeCheck = true
+
+		groupMembersRequest := group.BuildListUsersInGroupRequestWithNextToken(api.UserPoolId, nextToken)
+		groupMembersResponse, err := api.CognitoClient.ListUsersInGroup(groupMembersRequest)
+		if err != nil {
+			return nil, err
+		}
+
+		listOfUsers = append(listOfUsers, groupMembersResponse.Users...)
+		if groupMembersResponse.NextToken != nil {
```
daiLlew, line 117: The `else` isn't required - IMO less indentation is better/easier to read.
```
nextToken = ""
if groupMembersResponse.NextToken != nil {
    nextToken = *groupMembersResponse.NextToken
} 
```
This is personal opinion and I'm not requiring you to make this change - just offering my opinion.

models/groups.go
```diff
@@ -96,6 +97,22 @@ func (g *Group) BuildListUsersInGroupRequest(userPoolId string) *cognitoidentity
...
+			GroupName:  &g.Name,
+			UserPoolId: &userPoolId,
+		}
+	} else {
+		return &cognitoidentityprovider.ListUsersInGroupInput{
+			GroupName:  &g.Name,
+			UserPoolId: &userPoolId,
+			NextToken:  &nextToken,
+		}
+	}
```
daiLlew: The `else` not required here as there is a return inside the `if` block. Again no functional difference its just less code/indentation which I think makes it easier to read.
```
if nextToken == "" {
    return &cognitoidentityprovider.ListUsersInGroupInput{
	GroupName:  &g.Name,
	UserPoolId: &userPoolId,
}

return &cognitoidentityprovider.ListUsersInGroupInput{
    GroupName:  &g.Name,
    UserPoolId: &userPoolId,
    NextToken:  &nextToken,
}
```

models/groups_test.go
```diff
@@ -249,3 +250,34 @@ func TestGroup_BuildSuccessfulJsonResponse(t *testing.T) {
...
+			Name: "role-test",
+		}
+		userPoolId := "euwest-99-aabbcc"
+		nextToken := ""
+
+		response := group.BuildListUsersInGroupRequestWithNextToken(userPoolId, nextToken)
+
+		So(reflect.TypeOf(*response), ShouldEqual, reflect.TypeOf(cognitoidentityprovider.ListUsersInGroupInput{}))
+		So(*response.UserPoolId, ShouldEqual, userPoolId)
+		So(*response.GroupName, ShouldEqual, group.Name)
```
daiLlew, line 266: Add an assertion to verify that `NextToken` is nil/empty

HeJones: I did initially try with this ```So(*response.NextToken, ShouldBeNil)``` but I got the following error: ```runtime error: invalid memory address or nil pointer dereference ```. Do you know if there is a way to write a test without getting this error please?

HeJones: I think it works if I have `So(response.NextToken, ShouldBeNil)`

models/groups.go
HeJones: Thank you

api/groups.go
HeJones: Thank you

daiLlew (approved):

PR #69, opened 2021-08-11
update swagger for get users in a group endpoint

HeJones:
### What

- Update swagger spec to add the get users in a group endpoint
### How to review

Ensure the swagger spec reads

### Who can review

Anyone

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -386,6 +386,32 @@ paths:
           description: "Invalid request body or unexpected Cognito response"
           schema:
             $ref: '#/definitions/ErrorResponse'
+    get:
+      tags:
+        - Groups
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -386,6 +386,32 @@ paths:
...
+        - "application/json"
+      parameters:
+        - in: path
+          name: name
+          type: string
+          required: true
+          description: the group's name
+      responses:
+        200:
+          description: "User added to group"
```
mwnicks: Looks like this description has been copied from the post above, just wants to be updated to be relevant to this endpoint

HeJones: Thank you for spotting this, I have updated this.

mwnicks (approved):

PR #70, opened 2021-08-12
don't set the auth headers to test login response in develop

mwnicks:
### What

Temporarily disabled the auth headers to test if they are causing the nginx bug in develop

### How to review

Read the changes

### Who can review

Anyone but me


mike-kelly-methods (dismissed):

mike-kelly-methods (approved):
👍

PR #72, opened 2021-08-12
test with long headers not in JWT format

mwnicks:
### What

Testing long header values that are formatted differently than JWTs, to test the error is a length or format issue.

### How to review

Read the change

### Who can review

Anyone but me


mike-kelly-methods (approved):
👍

PR #78, opened 2021-08-16
test access token same length as in develop

mwnicks:
### What

Made the dummy test token equal in length to the ones created in develop

### How to review

Read the changes.

### Who can review

Anyone but me


mike-kelly-methods (approved):
👍

PR #81, opened 2021-08-19
Features/590 list groups for user

AWitcherONS:
### What

Add endpoint
GET /v1/user{id}/groups
ListGroupsForUser

### How to review

Ensure unit tests and feature tests pass
Ensure that code is to current standards
Postman tests 
[Identity-api ListGroupsForUser.postman_collection.txt](https://github.com/ONSdigital/dp-identity-api/files/7014519/Identity-api.ListGroupsForUser.postman_collection.txt)

### Who can review

Anyone except me


api/users.go
```diff
@@ -293,3 +293,53 @@ func (api *API) PasswordResetHandler(ctx context.Context, w http.ResponseWriter,
...
+	userID := models.UserParams{ID: vars["id"]}
+	listofgroupsInput := []*cognitoidentityprovider.GroupType{}
+	finalUserResponse := cognitoidentityprovider.AdminListGroupsForUserOutput{}
+	listusergroups := models.ListUserGroups{}
+
+	listofGroupsOutput, err := api.getGroupsForUser(listofgroupsInput, userID)
+	if err != nil {
+		cognitoErr := models.NewCognitoError(ctx, err, "Cognito ListofUserGroups request from list user groups endpoint")
+		if cognitoErr.Code == models.NotFoundError {
+			return nil, models.NewErrorResponse(http.StatusBadRequest, nil, cognitoErr)
```
daiLlew: Just checking if this is the right status code? Should it be bad request or not found?

cognito/mock/cognito_stub.go
```diff
@@ -559,3 +559,60 @@ func (m *CognitoIdentityProviderClientStub) DeleteGroup(input *cognitoidentitypr
...
+			NextToken: &nextTokenNil,
+		}, nil
+	}
+
+	if input.NextToken != nil && *input.NextToken != "" {
+		return &cognitoidentityprovider.AdminListGroupsForUserOutput{
+			Groups:    newGroups,
+			NextToken: &nextToken,
+		}, nil
+	} else {
```
daiLlew: `else` isn't required here as the logic will just fall through to this last return.

models/users.go
```diff
@@ -22,6 +22,12 @@ type UsersList struct {
 	Count int          `json:"count"`
 }
 
+type ListUserGroups struct {
```
daiLlew, line 40: Public types should have GoDoc.

models/users.go
```diff
@@ -22,6 +22,12 @@ type UsersList struct {
...
+	Groups    []*cognitoidentityprovider.GroupType `json:"Groups"`
+	NextToken *string                              `json:"NextToken"`
+	Count     int                                  `json:"Count"`
```
daiLlew: I think our JSON standards say that we should use lowercase with underscores e.g:
```
`json:"groups"`
`json:"next_token"`
`json:"count"`
```

models/users.go
```diff
@@ -276,6 +282,14 @@ type ListUsersOutput struct {
 	ListUsersOutput *cognitoidentityprovider.ListUsersOutput
 }
 
+type ListUserGroupsInput struct {
+	ListUserGroupsInput *cognitoidentityprovider.AdminListGroupsForUserInput
+}
+
+type ListUserGroupsOutput struct {
+	ListUserGroupsOutput *cognitoidentityprovider.AdminListGroupsForUserOutput
+}
```
daiLlew: GoDoc

models/users.go
```diff
@@ -449,3 +463,38 @@ func (p PasswordReset) BuildCognitoRequest(clientSecret string, clientId string)
...
+
+// BuildListUserGroupsRequest build the require input for cognito query to obtain the groups for given user
+func (p UserParams) BuildListUserGroupsRequest(userPoolId string, nextToken string) *cognitoidentityprovider.AdminListGroupsForUserInput {
+	if nextToken != "" {
+		return &cognitoidentityprovider.AdminListGroupsForUserInput{
+			UserPoolId: &userPoolId,
+			Username:   &p.ID,
+			NextToken:  &nextToken,
+		}
+	} else {
```
daiLlew: `else` not required - if the `if` condition is not true you can fall through to the other `return` statement. i.e.

```
if nextToken != "" {
    return &cognitoidentityprovider.AdminListGroupsForUserInput{
        UserPoolId: &userPoolId,
	Username:   &p.ID,
	NextToken:  &nextToken,
    }
}

return &cognitoidentityprovider.AdminListGroupsForUserInput{
    UserPoolId: &userPoolId,
    Username:   &p.ID,
}
```

models/users.go
```diff
@@ -449,3 +463,38 @@ func (p PasswordReset) BuildCognitoRequest(clientSecret string, clientId string)
...
+		return &cognitoidentityprovider.AdminListGroupsForUserInput{
+			UserPoolId: &userPoolId,
+			Username:   &p.ID,
+		}
+	}
+}
+
+//BuildListUserGroupsSuccessfulJsonResponse
```
daiLlew, line 500: Complete GoDoc.

models/users.go
```diff
@@ -449,3 +463,38 @@ func (p PasswordReset) BuildCognitoRequest(clientSecret string, clientId string)
...
+func (p *ListUserGroups) BuildListUserGroupsSuccessfulJsonResponse(ctx context.Context, result *cognitoidentityprovider.AdminListGroupsForUserOutput) ([]byte, error) {
+	userGroups := &ListUserGroups{}
+	userGroups.Groups = result.Groups
+	userGroups.NextToken = result.NextToken
+
+	if userGroups.Groups != nil {
+		userGroups.Count = len(result.Groups)
+	} else {
+		userGroups.Count = 0
+	}
```
daiLlew:
```
userGroups.Count  = 0
if userGroups.Groups != nil {
    userGroups.Count = len(result.Groups)
}
```
😄

cognito/mock/cognito_stub.go
AWitcherONS: One is built with a value in the nextToken 
the other is nextTokenNil (this is a badname because its an empty string) 
if not there then nextToken is null/nil which will cause aws cognito exception

models/users.go
```diff
@@ -449,3 +469,51 @@ func (p PasswordReset) BuildCognitoRequest(clientSecret string, clientId string)
...
+	return &cognitoidentityprovider.AdminListGroupsForUserInput{
+		UserPoolId: &userPoolId,
+		Username:   &p.ID}
+
+}
+
+//BuildListUserGroupsSuccessfulJsonResponse
+// formats the output to comply with current standards and to json , adds the count of groups returned and
+func (p *ListUserGroups) BuildListUserGroupsSuccessfulJsonResponse(ctx context.Context, result *cognitoidentityprovider.AdminListGroupsForUserOutput) ([]byte, error) {
+	for _, tmpGroup := range result.Groups {
```
daiLlew, line 508: As `result` is a pointer it can be `nil` so we should add a 
```
if result == nil {}
```
Check before we do the loop - if `result` is nil then `result.Groups` will error with nil pointer 

Simplified example: https://play.golang.org/p/Xps38s8iawz

AWitcherONS: Added the check and error return with corresponding unit tests

daiLlew (dismissed):

daiLlew (approved):

PR #82, opened 2021-08-19
Bug/list all users

mwnicks:
### What

Iterate through Cognito pagination responses to load all users in the list users endpoint rather than the 60 limited to a single Cognito response

### How to review

Call GET /v1/users, you should receive a 200 response with a body that contains a list of users and a count higher than 60.
(Assuming there are more than 60 users in the local user pool, which there are at the time of writing)

### Who can review

Anyone but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"time"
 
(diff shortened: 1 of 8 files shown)


mike-kelly-methods (approved):
👍

PR #92, opened 2021-09-06
Feature/upgrade logging v2

stmuk:
### What

Upgrade to logging v2

### How to review

Run tests & look at diff

### Who can review

Anyone


main_test.go
```diff
@@ -31,7 +32,7 @@ func (f *ComponentTest) InitializeScenario(ctx *godog.ScenarioContext) {
 
 	ctx.AfterScenario(func(*godog.Scenario, error) {
 		if err = component.Close(); err != nil {
-			log.Event(nil, "error closing identity component", log.Error(err), log.WARN)
+			log.Warn(context.Background(), "error closing identity component", log.FormatErrors([]error{err}))
```
redhug1, line 35: What is the reason for using:
log.FormatErrors([]error{err}
instead of just:
err
?

stmuk: I think it's an error caused by the automatic process I'm using.

I'll simplify it.

stmuk: The func signature is Warn(ctx context.Context, event string, opts ...option) and doesn't contain an error type.

redhug1: Ok.Thanks.

redhug1 (approved):
LGTM

PR #96, opened 2021-09-23
Feature/add list groups endpoint

AWitcherONS:
### What

added list groups endpoint 

### How to review

review code 
tests
and component tests 
Postman local get localhost:25600/v1/groups 
there are currently 214 groups ...

### Who can review
any one but me

### DIFF
--- a/api/groups.go
+++ b/api/groups.go
@@ -168,3 +168,54 @@ func (api *API) RemoveUserFromGroupHandler(ctx context.Context, w http.ResponseW
 
 	return models.NewSuccessResponse(jsonResponse, http.StatusOK, nil), nil
 }
+
+//List Groups for user pagination allows first call and then any other call if nextToken is not ""
+func (api *API) getListGroups() (*cognitoidentityprovider.ListGroupsOutput, error) {
+	firstTimeCheck := false
+	var nextToken string
+	group := models.ListUserGroupType{}
+
+	listOfGroups := cognitoidentityprovider.ListGroupsOutput{}
+	for {
+		if firstTimeCheck && nextToken == "" {
+			break
+		}
+		firstTimeCheck = true
+
+		listGroupsRequest := group.BuildListGroupsRequest(api.UserPoolId, nextToken)
+		listGroupsResponse, err := api.CognitoClient.ListGroups(listGroupsRequest)
+		if err != nil {
+			return nil, err
+		}
+
+		listOfGroups.Groups = append(listOfGroups.Groups, listGroupsResponse.Groups...)
+		nextToken = ""
+		if listGroupsResponse.NextToken != nil {
+			nextToken = *listGroupsResponse.NextToken
+		}
+	}
+	return &listOfGroups, nil
+}
+
+//ListGroupsHandler lists the users in the user pool
+func (api *API) ListGroupsHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+
+	// vars := mux.Vars(req)
+	finalGroupsResponse := models.ListUserGroups{}
+
+	listOfGroups, err := api.getListGroups()
+	if err != nil {
+		cognitoErr := models.NewCognitoError(ctx, err, "Cognito ListofUserGroups request from list user groups endpoint")
+		if cognitoErr.Code == models.NotFoundError {
+			return nil, models.NewErrorResponse(http.StatusNotFound, nil, cognitoErr)
+		}
+		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, cognitoErr)
+	}
+
+	jsonResponse, responseErr := finalGroupsResponse.BuildListGroupsSuccessfulJsonResponse(ctx, listOfGroups)
+	if responseErr != nil {
+		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, responseErr)
+	}
+	return models.NewSuccessResponse(jsonResponse, http.StatusOK, nil), nil
+
+}
--- a/api/groups.go
+++ b/api/groups.go
@@ -207,7 +207,7 @@ func (api *API) RemoveUserFromGroupHandler(ctx context.Context, w http.ResponseW
 	return models.NewSuccessResponse(jsonResponse, http.StatusOK, nil), nil
 }
 
-//List Groups for user pagination allows first call and then any other call if nextToken is not ""
+//List Groups pagination allows first call and then any other call if nextToken is not ""
 func (api *API) GetListGroups() (*cognitoidentityprovider.ListGroupsOutput, error) {
 	firstTimeCheck := false
 	var nextToken string
@@ -237,8 +237,6 @@ func (api *API) GetListGroups() (*cognitoidentityprovider.ListGroupsOutput, erro
 
 //ListGroupsHandler lists the users in the user pool
 func (api *API) ListGroupsHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
-
-	// vars := mux.Vars(req)
 	finalGroupsResponse := models.ListUserGroups{}
 
 	listOfGroups, err := api.GetListGroups()
--- a/api/groups.go
+++ b/api/groups.go
@@ -208,7 +208,7 @@ func (api *API) RemoveUserFromGroupHandler(ctx context.Context, w http.ResponseW
 }
 
 //List Groups for user pagination allows first call and then any other call if nextToken is not ""
-func (api *API) getListGroups() (*cognitoidentityprovider.ListGroupsOutput, error) {
+func (api *API) GetListGroups() (*cognitoidentityprovider.ListGroupsOutput, error) {
 	firstTimeCheck := false
 	var nextToken string
 	group := models.ListUserGroupType{}
@@ -241,7 +241,7 @@ func (api *API) ListGroupsHandler(ctx context.Context, w http.ResponseWriter, re
 	// vars := mux.Vars(req)
 	finalGroupsResponse := models.ListUserGroups{}
 
-	listOfGroups, err := api.getListGroups()
+	listOfGroups, err := api.GetListGroups()
 	if err != nil {
 		cognitoErr := models.NewCognitoError(ctx, err, "Cognito ListofUserGroups request from list user groups endpoint")
 		if cognitoErr.Code == models.NotFoundError {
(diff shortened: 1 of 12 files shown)


api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
 		}
 	})
 }
+
+func TestGetListGroups(t *testing.T) {
+	var ()
```
daiLlew: is this needed? Delete if not.

api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
...
+					{},
+				},
+			}
+			return listGroups, nil
+		}
+
+		listOfGroupsResponse, errorResponse := api.GetListGroups()
+
+		So(listOfGroupsResponse.Groups, ShouldResemble, listOfGroups)
+		So(len(listOfGroupsResponse.Groups), ShouldEqual, len(listOfGroups))
```
daiLlew:
```
So(listOfGroupsResponse.Groups, ShouldHaveLength, len(listOfGroups))
```

api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
 		}
 	})
 }
+
+func TestGetListGroups(t *testing.T) {
+	var ()
+
+	api, _, m := apiSetup()
+
+	Convey("When there is no next token cognito is called once and empty list of groups in returned", t, func() {
```
daiLlew: "When there is no next token cognito is called once and **an** empty list of groups in returned"

api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
...
+
+		m.ListGroupsFunc = func(input *cognitoidentityprovider.ListGroupsInput) (*cognitoidentityprovider.ListGroupsOutput, error) {
+			listGroups := &cognitoidentityprovider.ListGroupsOutput{
+				NextToken: nil,
+				Groups: []*cognitoidentityprovider.GroupType{
+					{},
+				},
+			}
+			return listGroups, nil
+		}
```
daiLlew: The next token is an integral part of this function so I think we should test & verify that it is calling out to Cognito the number of times we expect. 

I think we can do this with something similar to the following:  We create a count var outside the mock func and increment it inside each time its invoked. 
```
var count := 0

m.ListGroupsFunc = func(input *cognitoidentityprovider.ListGroupsInput) (*cognitoidentityprovider.ListGroupsOutput, error) {
                        count++
			listGroups := &cognitoidentityprovider.ListGroupsOutput{
				NextToken: nil,
				Groups: []*cognitoidentityprovider.GroupType{
					{},
				},
			}
			return listGroups, nil
		}
```
We can then verify the exact number of times the mock func was called proving the next token logic is correct and works as we expected.
```
So(count, ShouldEqual, 1)
```

api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
...
+
+		listOfGroupsResponse, errorResponse := api.GetListGroups()
+
+		So(listOfGroupsResponse.Groups, ShouldResemble, listOfGroups)
+		So(len(listOfGroupsResponse.Groups), ShouldEqual, len(listOfGroups))
+		So(listOfGroupsResponse.NextToken, ShouldBeNil)
+		So(errorResponse, ShouldBeNil)
```
daiLlew: Minor point - With Go we should always check the error response first. No harm here but it's a good habit to establish.

api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
...
+		)
+		listOfGroups := []*cognitoidentityprovider.GroupType{
+			{
+				Description: &description,
+				GroupName:   &group_name,
+				Precedence:  &precedence,
+			},
+		}
+
+		m.ListGroupsFunc = func(input *cognitoidentityprovider.ListGroupsInput) (*cognitoidentityprovider.ListGroupsOutput, error) {
```
daiLlew, line 947: See previous comment about tracking the number of invocations.

cognito/mock/cognito_stub.go
```diff
@@ -650,6 +650,28 @@ func (m *CognitoIdentityProviderClientStub) AdminListGroupsForUser(
 
 }
 
+func (m *CognitoIdentityProviderClientStub) ListGroups(
+	input *cognitoidentityprovider.ListGroupsInput) (*cognitoidentityprovider.ListGroupsOutput, error) {
```
daiLlew: I think this should be on the same line - Go isn't as fussy as Java on formatting long function signatures.

cognito/mock/cognito_stub.go
```diff
@@ -650,6 +650,28 @@ func (m *CognitoIdentityProviderClientStub) AdminListGroupsForUser(
...
+	nextToken := "nextToken"
+	nextTokenNil := ""
+
+	var (
+		newGroups []*cognitoidentityprovider.GroupType
+	)
```
daiLlew: This is unnecessary for declaring a single var scoped within a function. You can just do:

```
var newGroups []*cognitoidentityprovider.GroupType
```

features/steps/steps.go
```diff
@@ -145,3 +145,14 @@ func (c *IdentityComponent) thereAreRequiredNumberOfUsers(requiredNumberOfUsers
 	c.CognitoClient.AddMultipleUsers(requiredNumberOfUsersInt)
 	return nil
 }
+
+func (c *IdentityComponent) thereGroupsExistsInTheDatabase(groupCount string) error {
+
+	groupCountInt, err := strconv.Atoi(groupCount)
+	if err != nil {
+		return errors.New("could not convert group count to int")
```
daiLlew: Consider including the `groupCount` value in the error - always helpful to see what the value that failed to parse.

api/groups_test.go
```diff
@@ -895,3 +898,167 @@ func TestCreateNewGroup(t *testing.T) {
...
+				RoleArn:          aws.String(""),
+				UserPoolId:       aws.String(""),
+			},
+		}
+	)
+
+	api, w, m := apiSetup()
+
+	Convey("List groups -check expected responses", t, func() {
+		listGroupsTest := []struct {
```
daiLlew, line 1007: Consider adding a `Description string` to this test struct to hold a convey statement for each test case. In the test loop below then you can output the description and see which test case is running. 


```
for _, tt := range listGroupsTest {
    Convey(tt.Description, t, func() {
       // Existing test code here.
        ...
})
}
```
This will also provide some context for each of the test scenarios - which will help in future if we can any functionality and need to update the tests.

features/groups.feature
```diff
@@ -549,4 +551,33 @@ Feature: Groups
...
+                }
+            """  
+
+        Scenario: GET /v1/groups and checking the response status 200
+        Given there "100" groups exists in the database
+        When I GET "/v1/groups"
+        Then I should receive the following JSON response with status "200":
+            """
+                {
+                    "count":0,
```
daiLlew: Is this correct? I assume as groups exists this should be > 0?

models/groups.go
```diff
@@ -197,3 +197,54 @@ func (c *CreateGroup) GenerateGroupName() {
...
+			RoleArn:          tmpGroup.RoleArn,
+			UserPoolId:       tmpGroup.UserPoolId,
+		}
+
+		g.Groups = append(g.Groups, &newGroup)
+	}
+
+	g.NextToken = result.NextToken
+	g.Count = 0
+	if g.Groups != nil {
```
daiLlew: nil check here is not required `len(n)` will return 0 - see [Godoc](https://pkg.go.dev/builtin#len)

api/groups.go
```diff
@@ -206,3 +206,54 @@ func (api *API) RemoveUserFromGroupHandler(ctx context.Context, w http.ResponseW
...
+			nextToken = *listGroupsResponse.NextToken
+		}
+	}
+	return &listOfGroups, nil
+}
+
+//ListGroupsHandler lists the users in the user pool
+func (api *API) ListGroupsHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+
+	// vars := mux.Vars(req)
```
mike-kelly-methods: This can be removed.

api/groups_test.go
```diff
@@ -895,3 +898,241 @@ func TestCreateNewGroup(t *testing.T) {
...
+					So(responseBody.Groups, ShouldHaveLength, responseBody.Count)
+				},
+			},
+			{
+				"200 response from Cognito with populated NextToken",
+				"next_token",
+				func(input *cognitoidentityprovider.ListGroupsInput) (*cognitoidentityprovider.ListGroupsOutput, error) {
+					return &cognitoidentityprovider.ListGroupsOutput{
+						Groups:    groups,
+						NextToken: nil,
```
daiLlew, line 1071: I think this should be a non nil value for this test case?

api/groups_test.go
```diff
@@ -895,3 +898,241 @@ func TestCreateNewGroup(t *testing.T) {
...
+					So(errorResponse, ShouldBeNil)
+
+					So(successResponse.Status, ShouldEqual, http.StatusOK)
+					So(successResponse, ShouldNotBeNil)
+					So(successResponse.Body, ShouldNotBeNil)
+
+					var responseBody = models.ListUserGroups{}
+					json.Unmarshal(successResponse.Body, &responseBody)
+
+					So(responseBody.NextToken, ShouldBeNil)
```
daiLlew, line 1085: I think this should be `So(responseBody.NextToken, ShouldEqual, "next_token")`?

cognito/mock/cognito_stub.go
```diff
@@ -647,6 +648,20 @@ func (m *CognitoIdentityProviderClientStub) AdminListGroupsForUser(
 		Groups:    newGroups,
 		NextToken: &nextTokenNil,
 	}, nil
+}
+
+func (m *CognitoIdentityProviderClientStub) ListGroups(input *cognitoidentityprovider.ListGroupsInput) (*cognitoidentityprovider.ListGroupsOutput, error) {
+	fmt.Println(*input)
```
daiLlew: I assume this was adding for debugging purposes? If so remove it before committing or replace with a log call

api/groups_test.go
AWitcherONS: return NextToken will always be empty  but is used in the input   from the pagination loop will have no next_token 
awspagination

AWitcherONS: so the tests are faking an input with token....

daiLlew (approved):

PR #104, opened 2021-10-13
Add new update groups handler and associated logic.

mike-kelly-methods:
### What

As per title. Add new API endpoint to update a group's name (description) and precedence.

Group precedence is now `> 10` and `< 100`.

Add new API, Model and Feature tests to support functional logic.

### How to review

Tests and Build runs without errors. Check code and sanitise.

### Who can review

Anyone but me.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -91,6 +91,7 @@ func Setup(ctx context.Context, r *mux.Router, cognitoClient cognito.Client, use
 	r.HandleFunc("/v1/groups", contextAndErrors(api.ListGroupsHandler)).Methods(http.MethodGet)
 	r.HandleFunc("/v1/groups", contextAndErrors(api.CreateGroupHandler)).Methods(http.MethodPost)
 	r.HandleFunc("/v1/groups/{id}", contextAndErrors(api.GetGroupHandler)).Methods(http.MethodGet)
+	r.HandleFunc("/v1/groups/{id}", contextAndErrors(api.UpdateGroupHandler)).Methods(http.MethodPut)
 	r.HandleFunc("/v1/groups/{id}/members", contextAndErrors(api.AddUserToGroupHandler)).Methods(http.MethodPost)
 	r.HandleFunc("/v1/groups/{id}/members", contextAndErrors(api.ListUsersInGroupHandler)).Methods(http.MethodGet)
 	r.HandleFunc("/v1/groups/{id}/members/{user_id}", contextAndErrors(api.RemoveUserFromGroupHandler)).Methods(http.MethodDelete)
--- a/api/groups.go
+++ b/api/groups.go
@@ -21,12 +21,15 @@ func (api *API) CreateGroupHandler(ctx context.Context, w http.ResponseWriter, r
 	}
 
 	createGroup := models.CreateUpdateGroup{}
-	// GroupName is UUID
-	uuid := uuid.NewString()
-	createGroup.GroupName = &uuid
 	err = json.Unmarshal(body, &createGroup)
 	if err != nil {
 		return nil, handleBodyUnmarshalError(ctx, err)
+	}
+
+	// no groupname in body, set UUID
+	if createGroup.GroupName == nil {
+		uuid := uuid.NewString()
+		createGroup.GroupName = &uuid
 	}
 
 	createGroup.GroupsList, err = api.GetListGroups()
@@ -43,8 +46,6 @@ func (api *API) CreateGroupHandler(ctx context.Context, w http.ResponseWriter, r
 		return nil, models.NewErrorResponse(http.StatusBadRequest, nil, validationErrs...)
 	}
 
-	// remove special chars from description
-	createGroup.CleanGroupDescription()
 	// build create group input
 	input := createGroup.BuildCreateGroupInput(&api.UserPoolId)
 	_, err = api.CognitoClient.CreateGroup(input)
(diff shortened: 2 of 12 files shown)


api/groups_test.go
```diff
@@ -767,16 +789,46 @@ func TestCreateNewGroup(t *testing.T) {
 					So(errorResponse, ShouldBeNil)
 				},
 			},
-			// 400 response - group already exists
```
AWitcherONS, line 770: I don't understand why this test was removed 
surely that a test for whether a group already exists is valid 
and that response is more meaningful to a user...

AWitcherONS:
Shouldn't there be a swagger for PUT groups?
there were a few gofmt but nothing to fail 
test pass coverage of commit 
test-component passed

api/api.go
```diff
@@ -91,6 +91,7 @@ func Setup(ctx context.Context, r *mux.Router, cognitoClient cognito.Client, use
 	r.HandleFunc("/v1/groups", contextAndErrors(api.ListGroupsHandler)).Methods(http.MethodGet)
 	r.HandleFunc("/v1/groups", contextAndErrors(api.CreateGroupHandler)).Methods(http.MethodPost)
 	r.HandleFunc("/v1/groups/{id}", contextAndErrors(api.GetGroupHandler)).Methods(http.MethodGet)
+	r.HandleFunc("/v1/groups/{id}", contextAndErrors(api.UpdateGroupHandler)).Methods(http.MethodPut)
```
AWitcherONS, line 94: api_test.go should have unit test to cover new endpoint....
So(hasRoute(api.Router, "/v1/groups/{id}", http.MethodPut), ShouldBeTrue)

AWitcherONS (dismissed):

api/api.go
mike-kelly-methods: Added.

models/groups.go
```diff
@@ -148,39 +149,62 @@ func (g *Group) BuildSuccessfulJsonResponse(ctx context.Context) ([]byte, error)
...
 	var validationErrs []error
 
 	if g.Description == nil {
 		validationErrs = append(validationErrs, NewValidationError(ctx, InvalidGroupName, MissingGroupName))
 	} else if m, _ := regexp.MatchString("(?i)^role_.*", *g.Description); m {
 		validationErrs = append(validationErrs, NewValidationError(ctx, InvalidGroupName, IncorrectPatternInGroupName))
+	} else {
+		//ensure groups name in description doesn't already exist - creation only
+		if g.GroupsList != nil {
+			g.CleanGroupDescription()
```
mwnicks: Are we still cleaning the Group description?
I thought it was meant to be human readable now, as we have the UUID for urls etc.

api/groups.go
```diff
@@ -20,21 +20,32 @@ func (api *API) CreateGroupHandler(ctx context.Context, w http.ResponseWriter, r
...
+	}
+
+	validationErrs := createGroup.ValidateCreateUpdateGroupRequest(ctx)
 	if len(validationErrs) != 0 {
 		return nil, models.NewErrorResponse(http.StatusBadRequest, nil, validationErrs...)
 	}
 
-	// build create group input
+	// remove special chars from description
 	createGroup.CleanGroupDescription()
```
mwnicks: Isn't this already run in the validation step?
(although see comment in the validation step about if this is needed)

swagger.yaml
```diff
@@ -467,6 +467,46 @@ paths:
...
+      consumes:
+        - application/json
+      produces:
+        - "application/json"
+      parameters:
+        - in: body
+          name: "Group"
+          description: "The updated details of the group"
+          schema:
+            required: ["name", "precedence"]
```
mwnicks, line 489: Should this not be 'description'?

swagger.yaml
```diff
@@ -467,6 +467,46 @@ paths:
...
+            type: object
+            properties:
+              name:
```
mwnicks, line 492: Should this not be 'description'?

swagger.yaml
```diff
@@ -467,6 +467,46 @@ paths:
...
+    put:
+      tags:
+        - Groups
+      summary: "Update an existing group"
+      description: "Update an existing group in Cognito user pool"
+      consumes:
+        - application/json
+      produces:
+        - "application/json"
+      parameters:
```
mwnicks, line 479: Seems to be missing the `- in: path` parameter for the group name

api/groups.go
mike-kelly-methods: The `CleanGroupDescription()` function is applied to the incoming data after the validation step - as calling it before the validation, and without checking if the description is populated, would result in a null pointer exception.

models/groups.go
mike-kelly-methods: I wanted to double check this - thats fine, I can remove the regex replace on special charset then.

swagger.yaml
mike-kelly-methods: Changed this to `name` as per ticket https://trello.com/c/i7MqRBIg/670-670-add-human-readable-group-name:

`API Req -> Cognito - > API resp`
`name -> Description -> name`

swagger.yaml
mike-kelly-methods: As per above.

AWitcherONS (commented):
make tests pass 
make test-component pass
gofmt show format warnings but nothing to stop release
Postman suite for dp-identity-api run and all pass
no new Print statements

mwnicks (approved):
LGTM 👍

PR #108, opened 2021-10-18
update swagger for delete endpoint

Marianasilvago:
### What

Update swagger specifications for delete endpoint

### Who can review

Anyone.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -526,9 +526,9 @@ paths:
           required: true
           description: the group's name
       responses:
-        200:
+        204:
           description: "Deleted group "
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -526,7 +526,7 @@ paths:
           required: true
           description: the group's name
       responses:
-        200:
+        204:
           description: "Deleted group "
         400:
```
mike-kelly-methods: Should be `404` - StatusNotFound

mike-kelly-methods (approved):
👍

PR #109, opened 2021-10-19
Feature/add get user filter

AWitcherONS:
### What

Add user filter endpoint for active=<true/false>

### How to review
Check that code conforms to current standards
that make test run and the number of tests and coverage have increased
that make test-component run and the number of tests and coverage have increased
the following is a postman suit which includes the above 
[Identity-api.postman_collection.json.txt](https://github.com/ONSdigital/dp-identity-api/files/7374937/Identity-api.postman_collection.json.txt)

ensure that validation for the filter is in the appropriate place

### Who can review

Anyone except me

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -3,7 +3,6 @@ package api
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 
@@ -78,17 +77,19 @@ func (api *API) CreateUserHandler(ctx context.Context, w http.ResponseWriter, re
 //ListUsersHandler lists the users in the user pool
 func (api *API) ListUsersHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
 	var (
-		filterString          = aws.String("")
-		activeFilterString    = "AttributeName status \"True\""
-		notActiveFilterString = "AttributeName status \"False\""
+		filterString   = aws.String("")
+		validationErrs error
 	)
-	fmt.Printf(req.URL.RawQuery)
+
 	usersList := models.UsersList{}
-	if req.URL.RawQuery == "active=true" {
-		filterString = &activeFilterString
-	} else if req.URL.RawQuery == "active=false" {
-		filterString = &notActiveFilterString
+
+	if req.URL.RawQuery != "" {
+		*filterString, validationErrs = GetFilterStringAndValidate(req.URL.Path, req.URL.RawQuery)
+		if validationErrs != nil {
+			return nil, models.NewErrorResponse(http.StatusBadRequest, nil, validationErrs)
+		}
 	}
+
 	listUserResp, errResponse := api.ListUsersWorker(req.Context(), filterString, DefaultBackOffSchedule)
 	if errResponse != nil {
 		return nil, errResponse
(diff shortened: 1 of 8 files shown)


api/users.go
```diff
@@ -360,3 +373,26 @@ func (api *API) ListUserGroupsHandler(ctx context.Context, w http.ResponseWriter
...
+	}
+	for key, value := range filters {
+		if key == query && value["endpoint"] == endpoint {
+			return value["filterstring"], nil
+		}
+	}
+
+	return "", models.NewValidationError(ctx, models.InvalidFilterQuery, models.InvalidFilterQueryDescription)
+
+}
```
mike-kelly-methods, line 385:
```suggestion
func GetFilterStringAndValidate(path string, query string) (string, error) {
	ctx := context.Background()

	// struct of the form:
	//
	// "api_path": {
	//		{
	//			"query_param": "required_status",
	//			...,
	//		},
	//		{
	//			...,
	//			...,
	//		}
	// }
 	//
	filters := map[string]map[string]string{
		"/v1/users": {
			"active=true": "status=\"Enabled\"",
			"active=false": "status=\"Disabled\"",
		},
	}
	if filters[path] != nil && filters[path][query] != "" {
		return filters[path][query], nil
	} else {
		return "", models.NewValidationError(ctx, models.InvalidFilterQuery, models.InvalidFilterQueryDescription)
	}
}
```

mike-kelly-methods: I'd consider re-arranging `filters` struct to the above so that the api handler path is the main struct key - that way we can avoid having to loop through it.

We may want to build the `filters` struct outside the function so that we don't have to build it each time the `GetFilterStringAndValidate()` function is called.

A good place would be to add it to the `api.API{}` struct in the api package and name it something like `APIRequestFilter`:

	api := &API{
		Router:         r,
		CognitoClient:  cognitoClient,
		UserPoolId:     userPoolId,
		ClientId:       clientId,
		ClientSecret:   clientSecret,
		ClientAuthFlow: clientAuthFlow,
		AllowedDomains: allowedDomains,
		APIRequestFilter: map[string]map[string]string{
			"/v1/users": {
				"active=true": "status=\"Enabled\"",
				"active=false": "status=\"Disabled\"",
			},
		},
	}

mike-kelly-methods: Also, as discussed, perhaps moving it to a more suitable location in the codebase as to ensure that it's not tied to a particular package so that we can use this logic in any package if required in future iterations. (`../utilities/...`?)

features/users.feature
```diff
@@ -210,11 +210,33 @@ Feature: Users
...
+        Then the HTTP status code should be "200"
         And the list response should contain "70" entries
+    
+    Scenario: GET /v1/users?active=false with more than 60 active users and  70 inactive users checking the response status 200 with the correct number of users
+        Given there are "70" active users and "30" inactive users in the database
+        When I GET "/v1/users?active=false"
+        Then the HTTP status code should be "200"
+        And the list response should contain "30" entries
+
+    Scenario: GET /v1/users?active=anything with more than 60 active users and checking the response status 200 with the correct number of users
```
mike-kelly-methods: This scenario doesn't match the expected http status code of `400`?

features/users.feature
```diff
@@ -210,11 +210,33 @@ Feature: Users
...
+        Given there are "70" active users and "30" inactive users in the database
+        When I GET "/v1/users?active=anything"
+        Then the HTTP status code should be "400"
+    
+    Scenario: GET /v1/user?active=false with more than 60 active users and checking the response status 200 with the correct number of users
```
mike-kelly-methods: This scenario doesn't match the expected http status code of `404` - in fact, can't see anywhere a `404` is returned from the `ListUsersHandler` handler so maybe an invalid test?

mike-kelly-methods (approved):
👍

PR #121, opened 2021-12-13
Feature/restore

Marianasilvago:
### What

Import users, groups and group members from s3 to cognito

### How to review
Verify the code changes

### Who can review

Anyone.


scripts/groups/restore_groups.go
```diff
@@ -0,0 +1,137 @@
...
+		case "s3_bucket":
+			conf.s3Bucket = pair[1]
+		case "s3_base_dir":
+			conf.s3BaseDir = pair[1]
+		case "s3_region":
+			conf.s3Region = pair[1]
+		case "user_pool_id":
+			conf.awsCognitoUserPoolID = pair[1]
+		}
+	}
```
daiLlew: Can you update this to use the [github.com/kelseyhightower/envconfig](github.com/kelseyhightower/envconfig) library please? It will be consistent with the rest of our apps/libraries then. Here is an example of how to use that package:

https://github.com/ONSdigital/dp-identity-api/blob/e4c98e2179976d64526ab780da1f82025166001c/config/config.go#L12

scripts/groups/restore_groups.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		adduserToGroup(ctx, client, line, config.awsCognitoUserPoolID)
+		fmt.Println(line)
+	}
+	log.Info(ctx, "Successfully processed all the group members in S3 file")
+}
+
+func createGroup(line []string, client *cognito.CognitoIdentityProvider, ctx context.Context) {
+	createGroup := models.CreateUpdateGroup{}
+	createGroup.GroupName = &line[0]
```
daiLlew, line 106: We should probably check the line is the expected format - if not we can log and error and continue. As it is it will fail if the requested index does not exist and cause the whole process to bomb out.

I'd also suggest possibly tracking the line number and logging that out if there is an error. That way if we process 1000 entries and say 4 fail, we will know specifically which lines to investigate.

scripts/groups/restore_groups.go
```diff
@@ -0,0 +1,137 @@
...
+
+	_, err := client.CreateGroup(input)
+	if err != nil {
+		log.Error(ctx, fmt.Sprintf("failed to create group: %+v", createGroup), err)
+	}
+}
+
+func adduserToGroup(ctx context.Context, client *cognito.CognitoIdentityProvider, line []string, userPoolId string) {
+
+	userId := line[0]
```
daiLlew, line 124: Same here - check line format is as expected, track and log the line number in the event of an error.

scripts/users/restore_users.go
```diff
@@ -0,0 +1,89 @@
...
+		case "s3_region":
+			conf.s3Region = pair[1]
+		case "user_pool_id":
+			conf.awsCognitoUserPoolID = pair[1]
+		}
+	}
+	if conf.userFileName == "" {
+		fmt.Println("Please set Environment Variables ")
+		os.Exit(1)
+	}
```
daiLlew: See previous comment about config style.

pidah (approved):
LGTM

PR #126, opened 2022-01-27
renamed groupName -> ID & description -> Name

daiLlew:
### What

Updated the `group` entity renaming `name`-> `id` and `description` -> `name`.

Updated models, unit & component tests and swagger spec.


swagger.yaml
```diff
@@ -689,12 +696,12 @@ definitions:
...
+        type: string
       name:
         description: "The group name"
         type: string
-      description:
-        description: "A description of the group"
-        type: string
       precedence:
         description: "The priority of the group"
         type: integer
```
janderson2, line 707: Changes to the API are as expected, but while reviewing I noticed that we are declaring two definitions that are never used. If you wouldn't mind removing the following from the end of this spec file (line 839+) while you are here that would be much appreciated 🙂 

```

  AttributeType:
    description: "Specifies the attribute type standard or custom."
    type: object
    properties:
      Name:
        description: "The name of the attribute."
        type: string
      Value:
        description: "value of the attribute"
        type: string

  MFAOptionType:
    description: "SMS MFA configuration type"
    type: object
    properties:
      AttributeName:
        description: "The attribute name of the MFA option type. The only valid value is phone_number."
        type: string
      DeliveryMedium:
        description: "Parameter for SMS delivery medium value."
        type: string
```

models/error_content.go
```diff
@@ -61,7 +61,7 @@ const (
 	BodyReadFailedDescription              = "endpoint returned an error reading the request body"
 	InvalidPasswordDescription             = "the submitted password could not be validated"
 	PasswordGenerationErrorDescription     = "failed to generate a valid password"
-	MissingGroupNameErrorDescription       = "the group name was missing"
+	MissingGroupIDErrorDescription       = "the group ID was missing"
```
janderson2: Looks like a minor `go fmt` issue here.

models/groups.go
```diff
@@ -25,38 +25,38 @@ var (
 	groupPrecedenceMax    = int64(100)
 )
 
-//Type to map for the Cognito GroupType object
+//Group is a type to map for the Cognito GroupType object
```
janderson2: This wording confused me a little (and I realise this is existing wording here). This may just be me, but perhaps rewording to remove the mention of Cognito here could be more clear:

```suggestion
//Group is a type for the identity API representation of a group's details
```

I realise that this is also true of comments throughout the models that are not always clear that they are identity API models as opposed to Cognito API models due to references to Cognito. If we can avoid referencing Cognito here then that would be better since it may be that in the future we decide to swap out Cognito as the storage backend and implement our own using mongo for example so being clear what are API models means that we are clear on what the unchanging external interface to this microservice is.

models/groups.go
```diff
@@ -194,20 +194,20 @@ func (g *CreateUpdateGroup) ValidateCreateUpdateGroupRequest(ctx context.Context
...
 
 // BuildUpdateGroupInput builds a correctly populated UpdateGroupInput object using Groups values
 func (g *CreateUpdateGroup) BuildUpdateGroupInput(userPoolId string) *cognitoidentityprovider.UpdateGroupInput {
-	return &cognitoidentityprovider.UpdateGroupInput {
-		GroupName: g.GroupName,
-		Description: g.Description,
-		Precedence: g.Precedence,
-		UserPoolId: &userPoolId,
+	return &cognitoidentityprovider.UpdateGroupInput{
+		GroupName:   g.Name,
```
janderson2: I believe this line should likely be `g.ID`:

```suggestion
		GroupName:   g.ID,
```

swagger.yaml
```diff
@@ -615,10 +618,10 @@ paths:
         - "application/json"
       parameters:
         - in: path
-          name: name
+          name: id
           type: string
           required: true
-          description: the groups name
+          description: the groups ID
```
janderson2: Missing apostrophe:

```suggestion
          description: the group's ID
```

swagger.yaml
```diff
@@ -528,7 +531,7 @@ paths:
         - "application/json"
       parameters:
         - in: path
-          name: name
+          name: id
           type: string
           required: true
           description: the group's name
```
janderson2: Still referencing `name` rather than `id`:

```suggestion
          description: the group's ID
```

swagger.yaml
```diff
@@ -539,7 +542,7 @@ paths:
           description: "Invalid group name provided"
```
janderson2: Still referencing `name`:

```suggestion
          description: "Invalid group ID provided"
```

swagger.yaml
```diff
@@ -586,7 +589,7 @@ paths:
         - "application/json"
       parameters:
         - in: path
-          name: name
+          name: id
           type: string
           required: true
           description: the group's name
```
janderson2: Still referencing `name`:

```suggestion
          description: the group's ID
```

swagger.yaml
```diff
@@ -496,6 +496,9 @@ paths:
             required: ["name", "precedence"]
```
janderson2, line 496: Line 491 above is still referencing `name` when it should be `id`:

Current:
```
description: the group's name
```
Expected:
```
description: the group's ID
```

swagger.yaml
```diff
@@ -603,7 +606,7 @@ paths:
           description: "Unexpected Cognito response"
```
janderson2, line 606: Line 602 above still references `name` when it should reference `id`.

Current:
```
description: "Invalid group name provided"
```
Expected:
```
description: "Invalid group ID provided"
```

swagger.yaml
```diff
@@ -615,10 +618,10 @@ paths:
...
         - in: path
-          name: name
+          name: id
           type: string
           required: true
-          description: the groups name
+          description: the groups ID
         - in: path
           name: user_id
           type: string
```
janderson2, line 627: LIne 636 below still references group `name` rather than `id`:

Current:
```
description: "Invalid group name or user id provided"
```
Expected:
```
description: "Invalid group id or user id provided"
```

swagger.yaml
janderson2, line 707: `GroupsList` on line 716 has the field name for the array of groups as `users` (line 720) when it should be `groups`:

Current:
```
  GroupsList:
    description: "A list of groups for the response body"
    type: object
    properties:
      users:
        type: array
        items:
          $ref: '#/definitions/Group'
      count:
        type: integer
```
Expected:
```
  GroupsList:
    description: "A list of groups for the response body"
    type: object
    properties:
      groups:
        type: array
        items:
          $ref: '#/definitions/Group'
      count:
        type: integer
```

janderson2, line 707: Line 736 has an infinite recursion in the spec, but is not actually needed. We should delete this and update line 251 from `$ref: '#/definitions/UserGroups'` to `$ref: '#/definitions/GroupsList'`.

models/groups_test.go
```diff
@@ -534,12 +534,12 @@ func TestGroup_CreateUpdateGroupCleanGroupDescription(t *testing.T) {
 		)
```
janderson2, line 534: The name of this test function still references `description` rather than `name`:

Current:
```
func TestGroup_CreateUpdateGroupCleanGroupDescription(t *testing.T) {
```
Suggested:
```
func TestGroup_CreateUpdateGroupCleanGroupName(t *testing.T) {
```

swagger.yaml
janderson2, line 707: The enum on line 799 is not aligned to the list of possible values in [`error_content.go`](https://github.com/ONSdigital/dp-identity-api/blob/feature/rename-group-entity-fields/models/error_content.go#L10-L46).

janderson2 (commented):
The code changes themselves are mostly there, just one incorrect variable reference.

It looks like we haven't been the best at updating the swagger spec so I have add comments to address the parts that are incorrect. I recognise that this is scope creep, but it would be good to correct them if you wouldn't mind while we have noticed them.

Only other comments are minor.

swagger.yaml
daiLlew: I dont think that is right - having looking at the struct returned by this API endpoint it uses a different model? I could be wrong though. I will post my update so you can see what I think it should be.

daiLlew: Im not sure what you mean by this comment.

janderson2: You are right they are different 👍

janderson2: Sorry, that wasn't very clear. The list of possible enum values for the enum now on line 819 is incorrect and does not match the list of possible values which can be found in the code linked to in the comment above.

I think it should be (this is actually a subset of those shown since we appear to have codes that are types of internal server error so they should never go back to the user):

```
        enum:
          - "InvalidUserId"
          - "InvalidGroupID"
          - "InvalidForename"
          - "InvalidSurname"
          - "InvalidStatusNotes"
          - "InvalidEmail"
          - "InvalidToken"
          - "InternalServerError"
          - "NotFound"
          - "UserNotFound"
          - "GroupExists"
          - "GroupNotFound"
          - "DeliveryFailure"
          - "InvalidCode"
          - "ExpiredCode"
          - "InvalidField"
          - "InvalidPassword"
          - "LimitExceeded"
          - "NotAuthorised"
          - "PasswordResetRequired"
          - "TooManyFailedAttempts"
          - "TooManyRequests"
          - "UserNotConfirmed"
          - "UsernameExists"
          - "MissingConfig"
          - "UnknownRequestType"
          - "BodyCloseError"
          - "InvalidGroupName"
          - "InvalidGroupPrecedence"
          - "InvalidFilterQuery"
```

There is potentially a wider problem here of a need to review all of the error handling to replace any usages of specific codes for things that are internal server errors as these should always just get the `InternalServerError` code.

swagger.yaml
```diff
@@ -765,6 +765,824 @@ definitions:
...
+  title: dp-identity-api
+  description: "An API used to manage the authorisation of users accessing data publishing services."
+  version: 1.0.0
+  license:
+    name: "Open Government Licence v3.0"
+    url: "http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"
+basePath: "/v1"
+schemes:
+  - http
+paths:
```
janderson2: There seems to be something going on with this swagger spec as the header has been duplicated. Maybe a merge conflict issue?

janderson2 (approved):

PR #127, opened 2022-01-27
Mask internal server errors

Marianasilvago:
### What

Set  `Description` as `Internal Server Error` if the status code is `500`

### How to review

Make sure internal errors messages are not leaked.

### Who can review

Anyone.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -140,7 +140,7 @@ func writeErrorResponse(ctx context.Context, w http.ResponseWriter, errorRespons
 	}
 	w.WriteHeader(errorResponse.Status)
 	if errorResponse.Status == http.StatusInternalServerError {
-		jsonResponse, err = json.Marshal(models.Error{Code: models.InternalError, Description: "Internal Server Error"})
+		jsonResponse, err = json.Marshal(models.Error{Code: models.InternalError, Description: models.InternalErrorDescription})
 	} else {
 		jsonResponse, err = json.Marshal(errorResponse)
 	}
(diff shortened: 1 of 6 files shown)


api/api_test.go
```diff
@@ -195,6 +195,27 @@ func TestWriteErrorResponse(t *testing.T) {
...
+			},
+		}
+
+		resp := httptest.NewRecorder()
+
+		writeErrorResponse(ctx, resp, &errorResponse)
+
+		So(resp.Code, ShouldEqual, http.StatusInternalServerError)
+		So(resp.Result().Header.Get(WWWAuthenticateName), ShouldEqual, headerMsg)
+		So(resp.Body.String(), ShouldEqual, `{"code":"InternalServerError","description":"Internal Server Error"}`)
```
mike-kelly-methods: You can use the `models.InternalError` key here - as is being done in the api.go package above.

As per comment above, use the`"Internal Server Error"` description in the `API error descriptions` constants block here also.

api/api.go
```diff
@@ -137,8 +139,11 @@ func writeErrorResponse(ctx context.Context, w http.ResponseWriter, errorRespons
 		}
 	}
 	w.WriteHeader(errorResponse.Status)
-
-	jsonResponse, err := json.Marshal(errorResponse)
+	if errorResponse.Status == http.StatusInternalServerError {
+		jsonResponse, err = json.Marshal(models.Error{Code: models.InternalError, Description: "Internal Server Error"})
```
mike-kelly-methods: Rather than inlining, add the `"Internal Server Error"` description to the `API error descriptions` constants block and reference that.

mike-kelly-methods (approved):
👍

PR #128, opened 2022-03-04
Add JWKS parser endpoint - obtains JWT keys and associated RSA public…

mike-kelly-methods:
### What

As per title - Add JWKS parser endpoint to obtain JWT keys and associated RSA public signing keys.

### How to review

Run tests. Review code. Run service and hit the `/v1/jwt-keys` endpoint - will return associated key data for user pool.

### Who can review

!me


Marianasilvago (approved):
LGTM

PR #135, opened 2022-03-31
Retrieve JWKS RSA public keys config on startup and pass to the authorisation middleware

pidah:
### What

services that consume the authorisation middleware retrieve the JWKS RSA public keys via the identity-api JWKS handler endpoint. However, as this endpoint is not yet initialised for the identity-api itself during startup, a different process is needed to bootstrap the identity-api service with JWKS keys dynamically.
This change retrieves the JWKS RSA public keys from AWS Cognito during identity-api startup and passes it to the authorisation library middleware.

### How to review

Review the code and ensure the tests are ok

### Who can review

!me


Marianasilvago (dismissed):
LGTM

Marianasilvago (approved):

PR #149, opened 2022-06-06
Combine the DR restore scripts for users and teams into one script

Marianasilvago:
### What

Combine the DR restore scripts for users and teams into one script.

### How to review

Set required environment variable from readme file, and run import_users.go script.

### Who can review

Anyone.


pidah (approved):
LGTM

PR #153, opened 2022-06-22
Add message for sign out all user api

Marianasilvago:
### What

Return message with sign out all user api to inform user that action has been accepted.

### How to review

Ensure the code looks ok and tests pass.

### Who can review

Anyone.


AWitcherONS (approved):
I assume that something will happen in florence  because the screen had not changed but message in header

PR #169, opened 2022-09-20
add migration scripts

AWitcherONS:
### What

add aws cli scripts for cognito userpool migration and resilience 

### How to review

Describe the steps required to test the changes.

### Who can review

anyone but me

### DIFF
--- a/scripts/utils/migration_scripts/cognito_cleardown/README.md
+++ b/scripts/utils/migration_scripts/cognito_cleardown/README.md
@@ -1,0 +1,24 @@
+dp-identity-api Cognito Clear down
+================
+
+### Purpose
+
+These processes are to be used to make an existing cognito userpool back to a clean state.  
+If using an awsb environment you need to aws `sso login --profile <environment aws profile name>`  
+To view and amend execution permissions  
+`ls –l [file_name]`  
+`chmod +x filename`  
+
+Please run in the given order...
+
+### Step 1 remove all user team/group membership
+to run 
+`./delete_users_by_group.sh <userpool> <environment aws profile name>`
+
+### Step 2 remove all teams (except teams that are roles)
+to run
+`./delete_groups.sh <userpool> <environment aws profile name>`
+
+### Step 3 remove all users 
+to run
+`./delete_users.sh <userpool> <environment aws profile name>`
--- a/scripts/utils/migration_scripts/cognito_cleardown/README.md
+++ b/scripts/utils/migration_scripts/cognito_cleardown/README.md
@@ -4,7 +4,8 @@ dp-identity-api Cognito Clear down
 ### Purpose
 
 These processes are to be used to make an existing cognito userpool back to a clean state.  
-If using an awsb environment you need to aws `sso login --profile <environment aws profile name>`  
+If using an awsb environment you need to...  
+`aws sso login --profile <environment aws profile name>`  
 To view and amend execution permissions  
 `ls –l [file_name]`  
 `chmod +x filename`  
@@ -22,3 +23,10 @@ to run
 ### Step 3 remove all users 
 to run
 `./delete_users.sh <userpool> <environment aws profile name>`
+
+
+### Complete run...  
+`./delete_cognito_objects.sh
+<userpool> <environment aws profile name>`
+
+please note that these will continue if there is an issue in any  of the above steps 
--- a/scripts/utils/migration_scripts/cognito_cleardown/delete_cognito_objects.sh
+++ b/scripts/utils/migration_scripts/cognito_cleardown/delete_cognito_objects.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+user_pool_id=$1
+profile=${2:-default}
+
+chmod +x delete_users_by_group.sh
+chmod +x delete_groups.sh
+chmod +x delete_users.sh
+
+./delete_users_by_group.sh "$user_pool_id" "$profile"
+./delete_groups.sh "$user_pool_id" "$profile"
+./delete_users.sh  "$user_pool_id" "$profile"
+
+echo $(( $(date +%s) - start ))
--- a/scripts/utils/migration_scripts/cognito_cleardown/delete_cognito_objects.sh
+++ b/scripts/utils/migration_scripts/cognito_cleardown/delete_cognito_objects.sh
@@ -2,9 +2,7 @@
 user_pool_id=$1
 profile=${2:-default}
 
-chmod +x delete_users_by_group.sh
-chmod +x delete_groups.sh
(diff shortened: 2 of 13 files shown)


scripts/utils/migration_scripts/cognito_cleardown/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+These processes are to be used to make an existing cognito userpool back to a clean state.  
+If using an awsb environment you need to aws `sso login --profile <environment aws profile name>`  
+To view and amend execution permissions  
+`ls –l [file_name]`  
+`chmod +x filename`  
+
+Please run in the given order...
+
+### Step 1 remove all user team/group membership
```
nshumoogum, line 15: You could combine steps 1 to 3 in make target, therefore developer less likely to make a mistake

scripts/utils/migration_scripts/cognito_cleardown/delete_groups.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+user_pool_id=$1
+profile=${2:-default}
+
+roles=("role-admin" "role-publisher")
```
nshumoogum, line 6: Is it worth leaving a comment above this to say `# roles to keep`

nshumoogum (dismissed):

nshumoogum:
Two suggestions, if you do implement them, then let me know and I can re-review

scripts/utils/migration_scripts/collectionsMigration/Makefile
```diff
@@ -0,0 +1,49 @@
...
+GOOS?=$(shell go env GOOS)
+GOARCH?=$(shell go env GOARCH)
+
+BUILD=build
+BUILD_ARCH=$(BUILD)/$(GOOS)-$(GOARCH)
+BUILD_SCRIPT=$(BUILD_ARCH)/$(APP).sh
+
+.PHONY: all
+all: clean
+	make GOOS=linux build deploy
```
jessjenkins: Why is `make all` setting `GOOS=linux`? Surely the os should be `darwin` when run locally. Only deployments to AWS should be `linux`. 
On a related note, it's not a good idea to set the `GOARCH` to the local architecture especially as M1 macs become more prevalent. In those cases the arch will become `arm64` rather than the `amd64` required by our current AWS deployments.

scripts/utils/migration_scripts/collectionsMigration/main.go
```diff
@@ -0,0 +1,219 @@
...
+	collectionCopyDir string
+	migration bool
+}
+
+type teamCollection struct {
+	id   string
+	name string
+}
+
+type collection struct {
```
jessjenkins: How can we be sure this is the complete list of fields? Any fields not defined in this list will be lost on Marshalling into the struct and then won't be left in the resulting json file. I suggested we should use jq for modifying the files originally as it would preserve all other fields and only modify the one field we wanted to change. I'd still suggest doing this now even if we use something like `interface{}` within go and then iterate to find the field rather than Marshalling the struct explicitly like this.

scripts/utils/migration_scripts/collectionsMigration/Makefile
AWitcherONS: added local running instructions and comment that the make file is about running remotely

scripts/utils/migration_scripts/collectionsMigration/main.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, f := range tmpdirFilesList {
+		if strings.Contains(f.Name(), ".json") {
+			dirFilesList = append(dirFilesList, f)
+		}
+	}
+
+	return dirFilesList, path, nil
+}
+
+func getTeamsList(dir string) []teamCollection {
```
jessjenkins: Instead of returning a slice of teams, you could instead return two maps `(teamids,teamnames map[string]string`). 
This would make the the lookups much simpler in the amend/revert functions. (see comments below)

scripts/utils/migration_scripts/collectionsMigration/main.go
```diff
@@ -0,0 +1,219 @@
...
+		var result collection
+		_ = json.Unmarshal(collectionFile, &result)
+		for ind, collectionTeam := range result.Teams {
+			for _, team := range teamsList {
+				if team.name == collectionTeam {
+					result.Teams[ind] = team.id
+					*saveFile = true
+				}
+			}
+		}
```
jessjenkins: If you use a map for teams you could use something like…
```
newid, ok := teamids[collectionTeam]
if ok {
 result.Teams[ind] = newid
}
```

scripts/utils/migration_scripts/collectionsMigration/main.go
```diff
@@ -0,0 +1,177 @@
...
+			strName = name
+		}
+		teamsList[strID] = strName
+
+	}
+
+	return teamsList
+}
+
+func getTeamsListName(dir string) map[string]string {
```
jessjenkins: this function is almost identical to the one above. It'd be better to have a single function that returns two reciprocal maps (ids[name], names[id]).

scripts/utils/migration_scripts/collectionsMigration/main.go
```diff
@@ -0,0 +1,177 @@
...
+	teamsList := make(map[string]string)
+	var result map[string]interface{}
```
jessjenkins: We could have kept the struct for the teams files as we are not writing them back. It was only the collections files that needed `map[string]interface{}`

scripts/utils/migration_scripts/collectionsMigration/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	make GOOS=linux build deploy
```
jessjenkins: This line is a duplicate of the line above.

jessjenkins (commented):
Reviewed everything in `collectionsMigration` folder and I am happy with it. Might be worth getting @nshumoogum to re-review the other scripts for his requested changes.

scripts/utils/migration_scripts/cognito_cleardown/delete_cognito_objects.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+user_pool_id=$1
+profile=${2:-default}
+
+chmod +x delete_users_by_group.sh
+chmod +x delete_groups.sh
+chmod +x delete_users.sh
```
jessjenkins: No need to chmod the scripts. Git saves the file mode of them anyway.

jessjenkins (approved):

PR #185, opened 2023-04-20
Feature/compose auth stack

lindenmckenzie:
### What

Altered local docker setup to work with dp-compose auth stack: https://github.com/ONSdigital/dp-compose/pull/126

I also removed the generation of the Cognito secrets via the awscli in `make debug` as the permissions are not present anymore to do this in AWSB. 

### How to review

Use https://github.com/ONSdigital/dp-compose/pull/126 to run local auth stack and try to login to Florence with sandbox credentials

### Who can review

Not me.


jondewijones (approved):
LGTM

PR #194, opened 2024-03-07
Add "Teams Report 

AWitcherONS:
### What

Teams Report 
to produce a list of all groups and th email of members of the groups

### How to review

code confirms to current standards
lint and audit comply
all test pass

### Who can review
anyone but me


api/groups.go
```diff
@@ -484,3 +484,64 @@ func (api *API) RemoveUserFromGroup(ctx context.Context, group models.Group, use
 
 	return &listOfUsers, nil
 }
+
+func (api *API) ListGroupsUsersHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	var (
+		GroupsUsersList []models.ListGroupUsersType
+		header1         = "GroupName"
+		header2         = "UserEmail"
```
lindenmckenzie: Would be useful for these to have more meaningful names.

api/groups.go
```diff
@@ -484,3 +484,64 @@ func (api *API) RemoveUserFromGroup(ctx context.Context, group models.Group, use
 
 	return &listOfUsers, nil
 }
+
+func (api *API) ListGroupsUsersHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
```
lindenmckenzie, line 491: Should have a GoDoc here.

api/groups.go
```diff
@@ -484,3 +484,64 @@ func (api *API) RemoveUserFromGroup(ctx context.Context, group models.Group, use
...
+			for _, attribute := range user.Attributes {
+				if *attribute.Name == "Email" {
+					GroupsUsersList = append(GroupsUsersList, models.ListGroupUsersType{
+						GroupName: *ListGroup.Description,
+						UserEmail: *attribute.Value,
+					})
+				}
+			}
+		}
+	}
```
lindenmckenzie: Would suggest separating this into a different function

api/groups.go
```diff
@@ -484,3 +484,64 @@ func (api *API) RemoveUserFromGroup(ctx context.Context, group models.Group, use
...
+		rows := [][]string{}
+		rows = append(rows, []string{header1, header2})
+
+		for _, record := range GroupsUsersList {
+			rows = append(rows, []string{record.GroupName, record.UserEmail})
+		}
+
+		w.WriteAll(rows)
+		return models.NewSuccessResponse(buf.Bytes(), http.StatusOK, nil), nil
+	}
```
lindenmckenzie: Could also separate this into different function.

api/groups.go
```diff
@@ -484,3 +484,64 @@ func (api *API) RemoveUserFromGroup(ctx context.Context, group models.Group, use
...
+	for _, ListGroup := range listOfGroups.Groups {
+
+		inputGroup := models.Group{ID: *ListGroup.GroupName}
+		var listOfUsersInput []*cognitoidentityprovider.UserType
+		listUsers, err := api.getUsersInAGroup(listOfUsersInput, inputGroup)
+		if err != nil {
+			cognitoErr := models.NewCognitoError(ctx, err, "Cognito ListUsersInGroup request from list users in group endpoint")
+			if cognitoErr.Code == models.NotFoundError {
+				return nil, models.NewErrorResponse(http.StatusNotFound, nil, cognitoErr)
+			}
```
lindenmckenzie: Tracking this logic through - if we: 
- get a list of all groups 
- request the users for each group
- a single group request returns a 404
- we return a 404 to the client/user for the overall response

Would seem like a server error is more appropriate here as something's gone wrong in processing those groups. Otherwise this could be conflated with their being no groups, as per line 499.

api/groups_test.go
```diff
@@ -28,14 +31,14 @@ const (
...
 
 func TestAddUserToGroupHandler(t *testing.T) {
 	var (
-		ctx           = context.Background()
-		userId string = "abcd1234"
+		ctx    = context.Background()
+		userId = "abcd1234"
 
-		userNotFoundDescription, groupNotFoundDescription, internalErrorDescription string = "user not found", "group not found", "internal error"
+		userNotFoundDescription, groupNotFoundDescription, internalErrorDescription = "user not found", "group not found", "internal error"
```
lindenmckenzie: Why are we declaring them as defaults and then overriding to the exact same value?

lindenmckenzie: Same question for other examples lower down.

api/groups_test.go
```diff
@@ -1547,7 +1548,7 @@ func TestSetGroupUsersHandler(t *testing.T) {
 			RoleArn:          aws.String(""),
 			UserPoolId:       aws.String("")}
 		// userNotFoundDescription, groupNotFoundDescription, internalErrorDescription string = "user not found", "group not found", "internal error"
```
lindenmckenzie: Why's this comment here?

api/groups_test.go
```diff
@@ -2351,5 +2352,324 @@ func TestAddUserToGroup(t *testing.T) {
...
 	})
+}
+
+func TestListGroupsUsersHandler(t *testing.T) {
+	var (
+		ctx = context.Background()
+	)
+
+	api, w, m := apiSetup()
+	Convey("check expected responses json", t, func() {
```
lindenmckenzie: For these Convey steps - it's useful to list it out like you're doing BDD to help improve cognition. 

For example:

when the listGroupsUsersHandler is called
and cognito returns 0 groups successfully
then it should return an empty json response
and it should return a csv with just the header rows

api/groups_test.go
```diff
@@ -2351,5 +2352,324 @@ func TestAddUserToGroup(t *testing.T) {
...
+		groupList = append(groupList, &groups)
+	}
+	output := cognitoidentityprovider.ListGroupsOutput{
+		Groups:    groupList,
+		NextToken: nil,
+	}
+	return output
+}
+
+func ListGroupsUsers(noOfUsers int) *cognitoidentityprovider.ListUsersInGroupOutput {
```
lindenmckenzie: Does this need to be a public function?

api/groups_test.go
```diff
@@ -2351,5 +2352,324 @@ func TestAddUserToGroup(t *testing.T) {
...
+				func(successResponse *models.SuccessResponse, errorResponse *models.ErrorResponse) {
+					var testOutJSON []models.ListGroupUsersType
+					jsonErr := json.Unmarshal(successResponse.Body, &testOutJSON)
+					So(jsonErr, ShouldBeNil)
+					So(testOutJSON, ShouldNotBeEmpty)
+					So(testOutJSON, ShouldHaveLength, 1)
+					for _, t := range testOutJSON {
+						So(t.GroupName, ShouldResemble, "group 0 description")
+						So(t.UserEmail, ShouldResemble, "user0.email@domain.test")
+					}
```
lindenmckenzie: why aren't you testing like this in the CSV ones lower down?

api/groups_test.go
```diff
@@ -2351,5 +2352,324 @@ func TestAddUserToGroup(t *testing.T) {
...
+				func(successResponse *models.SuccessResponse, errorResponse *models.ErrorResponse) {
+					var testOutJSON []models.ListGroupUsersType
+					jsonErr := json.Unmarshal(successResponse.Body, &testOutJSON)
+					So(jsonErr, ShouldBeNil)
+					So(testOutJSON, ShouldNotBeEmpty)
+					So(testOutJSON, ShouldHaveLength, 6)
+					testOutCSV := string(successResponse.Body[:])
+					stringSlice := strings.Split(testOutCSV, "\n")
+					So(stringSlice, ShouldHaveLength, 1)
+					So(stringSlice[0], ShouldNotBeEmpty)
```
lindenmckenzie: I'm not sure we need the additional tests to check it's not in csv format if we've successfully unmarshalled the json.

api/groups_test.go
lindenmckenzie: this is a general comment around all of these tests below.

api/groups.go
lindenmckenzie: The 404s don't have any test coverage either.

api/groups_test.go
lindenmckenzie: GoDoc would be useful here

api/groups_test.go
```diff
@@ -2351,5 +2352,324 @@ func TestAddUserToGroup(t *testing.T) {
...
+
+func ListGroupsUsers(noOfUsers int) *cognitoidentityprovider.ListUsersInGroupOutput {
+	userList := []*cognitoidentityprovider.UserType{}
+	var (
+		attributeEmail = "Email"
+	)
 
+	for i := 0; i < noOfUsers; i++ {
+		userAttributes := []*cognitoidentityprovider.AttributeType{}
+		userEmail := "user" + strconv.Itoa(i) + ".email@domain.test"
```
lindenmckenzie: fmt.Sprintf might be a better choice here.

lindenmckenzie: could combine it with the Username lower down as well. So then Username and Email are both using the same base pattern. E.g. 

Username: "user_0"
UserEmail: "user_0.email@domain.test"

features/groups.feature
```diff
@@ -1,6 +1,6 @@
 Feature: Groups
 
-#   Create new group scenarios
+#   Create new group goscenarios
```
lindenmckenzie: Why this change?

features/groups.feature
```diff
@@ -895,3 +898,29 @@ Scenario: PUT /v1/groups/{id}/members and checking the response status 200
         """
     Then the HTTP status code should be "404"
 
+#   Get getgroupsreport scenarios
```
lindenmckenzie: You could use Gherkin tags rather than comments to split scenarios.

features/groups.feature
```diff
@@ -895,3 +898,29 @@ Scenario: PUT /v1/groups/{id}/members and checking the response status 200
...
+                      "group": "test-group description",
+                      "user": "email@ons.gov.uk"
+                    },
+                    {
+                      "group": "test-group2 description",
+                      "user": "otheremail@ons.gov.uk"
+                    }
+                  ]
+            """
+
```
lindenmckenzie: Missing a CSV request or any unhappy paths.

features/steps/steps.go
```diff
@@ -241,3 +250,27 @@ func (c *IdentityComponent) theResponseShouldMatchTheFollowingJsonForListgroups(
...
+		return
+	}
+
+	responseBody := c.apiFeature.HTTPResponse.Body
+	resBody, _ := io.ReadAll(responseBody)
+
+	if err = json.Unmarshal(resBody, &actual); err != nil {
+		return
+	}
+	// the matching may be adapted per different requirements.
```
lindenmckenzie: what does this comment mean?

features/steps/steps.go
```diff
@@ -241,3 +250,27 @@ func (c *IdentityComponent) theResponseShouldMatchTheFollowingJsonForListgroups(
 	}
 	return nil
 }
+
+func (c *IdentityComponent) theResponseShouldMatchTheFollowingJsonForListgroupsusers(body *godog.DocString) (err error) {
```
lindenmckenzie: There is a step in `dp-component-test` you could use for this:

`I should receive the following JSON response:`

https://github.com/ONSdigital/dp-component-test/blob/6f7c3b28d792614796490189bacc3458f5db487d/api_feature.go#L64

lindenmckenzie: it's more generic so more re-usable.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/ONSdigital/dp-identity-api
 
 go 1.21
 
+replace google.golang.org/protobuf => google.golang.org/protobuf v1.33.0
```
lindenmckenzie, line 5: It's good practice to only upgrade the specific version that we're calling to ensure that we then don't downgrade by accident later.

lindenmckenzie: Odd this hasn't shown up on the go.sum though.

models/groups.go
```diff
@@ -25,6 +25,15 @@ var (
 	groupPrecedenceMax    = int64(100)
 )
 
+// ListGroupUsersType list of groups and the membership returning group description and user email
+type ListGroupsUsers struct {
+	Groups []*ListGroupUsersType `json:"groups"`
+}
+type ListGroupUsersType struct {
```
lindenmckenzie, line 29: I'm not sure on how clear on the struct name is, could you explain further?

swagger.yaml
```diff
@@ -79,7 +79,7 @@ paths:
     put:
       tags:
         - Tokens
-      summary: "Refreshs a users access token using the refresh and ID tokens"
+      summary: "Refreshes a users access token using the refresh and ID tokens"
```
lindenmckenzie, line 82: 👍 👍 👍 👍 👍

swagger.yaml
```diff
@@ -351,7 +351,7 @@ paths:
           schema:
             $ref: '#/definitions/PasswordResetExpirationTime'
         400:
-          description: "Unable to validate the type, email, passowrd or session in the request"
+          description: "Unable to validate the type, email, password or session in the request"
```
lindenmckenzie, line 354: 👍 👍 👍 👍 👍 👍

swagger.yaml
```diff
@@ -447,6 +447,35 @@ paths:
...
             $ref: '#/definitions/ErrorResponse'
+  /groups-report:
+    get:
+      tags:
+        - Groups
+      summary: "List groups and the membership of the groups"
+      description: "
+      If a preview team ( cognito group ) is empty of users then this is not on the report.  
+      If the preview team ( cognito group ) acts as a role (i.e. publishers, admin) then this is included on the report
+      if header accept "
```
lindenmckenzie: This seems to end half way through a sentence?

swagger.yaml
```diff
@@ -447,6 +447,35 @@ paths:
...
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
```
lindenmckenzie: think you can add `text/csv` to this array here instead of the header parameter below.

swagger.yaml
```diff
@@ -773,6 +802,22 @@ definitions:
...
+    type: object
+    properties:
+      groups:
+        type: array
+        items:
+          $ref: '#/definitions/Group'
+  GroupMember:
+    description: "An entry of group description and the membership user email"
+    type: object
+    properties:
```
lindenmckenzie: I'm not entirely certain of the descriptions here, could you explain further?

swagger.yaml
```diff
@@ -447,6 +447,35 @@ paths:
...
+          description: if text/csv will return csv list else json
+      responses:
+        200:
+          description: "The list of groups and members"
+          schema:
+            $ref: '#/definitions/GroupsUsersList'
+        500:
+          description: "Internal error occurred"
+          schema:
+            $ref: '#/definitions/ErrorResponse'
```
lindenmckenzie: In the code there's possibilities of a 404 at the moment (which I'm not sure about - see groups.go)

swagger.yaml
```diff
@@ -447,6 +447,35 @@ paths:
...
+      if header accept "
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: header
+          name: Accept
+          type: string
+          required: true
```
lindenmckenzie: if you do keep this it's not strictly speaking required if we have default

lindenmckenzie (commented):
Few comments on here - happy to chat through. Really strong start though and great tidying up of some other linting/typos etc throughout the rest of the application.

cognito/mock/cognito_stub.go
```diff
@@ -534,12 +553,19 @@ func (m *CognitoIdentityProviderClientStub) ListUsersInGroup(input *cognitoident
 		}
 		userList = append(userList, &userDetails)
 	}
+
+	//fmt.Println("----- ListGroupUsers output ", cognitoidentityprovider.ListUsersInGroupOutput{
+	//	Users: userList,
+	//})
+
```
lindenmckenzie: Why is this being left in?

features/groups.feature
```diff
@@ -488,410 +579,450 @@ Scenario: DELETE /v1/groups/{id}/members/{user_id} and checking the response sta
...
-            {"code":"InternalServerError", "description":"Internal Server Error"}
-        """
+            """
+        Then the HTTP status code should be "404"
+
+@groups-report
+    Scenario: GET /v1/groups-report checking the response status 200 got an empty report no groups
+        Given I am an admin user
+        When I GET "/v1/groups-report"
+    Then the response code should be 200
```
lindenmckenzie: Formatting here.

lindenmckenzie: Should you be suggesting an expected response body here as well?

features/groups.feature
```diff
@@ -488,410 +579,450 @@ Scenario: DELETE /v1/groups/{id}/members/{user_id} and checking the response sta
...
+
+@groups-report
+    Scenario: GET /v1/groups-report checking the response status 200 an empty report with one groups but no users
+        Given I am an admin user
+        And group "test-group" and description "test group description" exists in the database
+        And a user with username "abcd1234" and email "email@ons.gov.uk" exists in the database
+        When I GET "/v1/groups-report"
+        Then the response code should be 200
```
lindenmckenzie, line 917: Should you be suggesting an expected response body here as well?

features/groups.feature
```diff
@@ -488,410 +579,450 @@ Scenario: DELETE /v1/groups/{id}/members/{user_id} and checking the response sta
...
+                        {"group": "test group_2 description",
+                            "user": "email2@ons.gov.uk"},
+                        {"group": "test group_4 description",
+                            "user": "email1@ons.gov.uk"},
+                        {"group": "test group_4 description",
+                            "user": "email3@ons.gov.uk"},
+                        {"group": "test group_5 description",
+                            "user": "email1@ons.gov.uk"},
+                        {"group": "test group_5 description",
+                            "user":  "email3@ons.gov.uk" }]
```
lindenmckenzie, line 985: This JSON is reflecting the needs of the CSV rather than how it might be structured in JSON instead. Which could be:
```
[
   {
       "group": "test group_1 description",
       "users": [
           {
               "email": "email1@ons.gov.uk"
           },
           {
               "email": "email2@ons.gov.uk"
           }
       ]
   }
]
```

swagger.yaml
```diff
@@ -447,6 +447,58 @@ paths:
...
+            application/json:
+              [
+                {
+                  "group": "group dexription",
+                  "user": "user.email@emaildomain"
+                }
+              ]
+            text/csv:
+              "group, user 
+              group dexription,user.email@emaildomain
```
lindenmckenzie:
```suggestion
              group description,user.email@emaildomain
```

swagger.yaml
```diff
@@ -447,6 +447,58 @@ paths:
...
+      responses:
+        200:
+          description: OK The list of groups and members
+          schema:
+            $ref: '#/definitions/GroupsUsersList'
+          examples:
+            application/json:
+              [
+                {
+                  "group": "group dexription",
```
lindenmckenzie:
```suggestion
                  "group": "group description",
```

swagger.yaml
```diff
@@ -447,6 +447,58 @@ paths:
...
+        - Groups
+      summary: "List groups and the membership of the groups"
+      description: " To return a list of :-
+          preview team ( cognito group ) description
+          preview team ( cognito group ) member user email
+          json by default 
+          or if Request Header Accept 'text/csv' then csv format, consisting of preview team description amd user email. 
+        Exceptions :- 
+          preview team ( cognito group ) is empty of members(users).  
+          preview team ( cognito group ) acts as a role (i.e. publishers, admin) then this is included on the report"
```
lindenmckenzie: I thought it wasn't included in the report in this instance?

lindenmckenzie (commented):
Few minor comments and some things to think about (potentially for the future)

swagger.yaml
```diff
@@ -447,6 +447,60 @@ paths:
...
+      consumes:
+        - "application/json"
+      parameters:
+        - in: header
+          name: Accept:application/json
+          type: string
+          required: false
+          description: "Produce array of json"
+        - in: header
+          name: Accept:text/scv
```
lindenmckenzie:
```suggestion
          name: Accept:text/csv
```

lindenmckenzie (approved):

PR #202, opened 2024-04-24
Feature/add sort groups

andre-urbani:
### What

Add functionality to sort groups by name, ascending and descending order
Trello - https://trello.com/c/b5yI8VPX/2118-add-sort-parameter-to-groups-endpoint

### How to review

Changes make sense, tests pass

### Who can review

Anyone


api/groups.go
```diff
@@ -24,6 +26,7 @@ const (
 
 // CreateGroupHandler creates a new group
 func (api *API) CreateGroupHandler(ctx context.Context, w http.ResponseWriter, req *http.Request) (*models.SuccessResponse, *models.ErrorResponse) {
+	sortBy := req.URL.Query().Get("sortBy")
```
gedge: can you coordinate with @dionisislolas  about what parameter to use please (`sortBy` or `sort`)

api/groups.go
```diff
@@ -601,3 +612,49 @@ func (api *API) GetTeamsReportLines(listOfGroups *cognitoidentityprovider.ListGr
...
+		case "desc":
+			sortByGroupName(groups, false)
+			return
+		default:
+			dplogs.Info(ctx, "groups.sortGroups: Not a correct sort by value. Groups not sorted.", dplogs.Data{"sort": sortBy})
+			return
+		}
+	default:
+		dplogs.Info(ctx, "groups.sortGroups: Not a correct sort by value. Groups not sorted.", dplogs.Data{"sort": sortBy})
+		return
```
lindenmckenzie: shouldn't we be returning a 400 in this instance?

lindenmckenzie: what happens if I put in /groups?sort=dogs:asc

api/groups.go
```diff
@@ -601,3 +612,49 @@ func (api *API) GetTeamsReportLines(listOfGroups *cognitoidentityprovider.ListGr
...
+	}
+}
+
+func sortByGroupName(groups []*cognitoidentityprovider.GroupType, ascending bool) {
```
lindenmckenzie, line 647: godoc etc.

api/groups_test.go
```diff
@@ -3065,3 +3065,72 @@ func listGroupsUsersMock(noOfGroups, noOfUsers int) ([]models.ListGroupUsersType
...
+		})
+		Convey("When sorting with an invalid asc or desc", func() {
+			sort := strings.Split("name:xyz", ":")
+			sortGroups(ctx, &groups, sort)
+			Convey("The groups should remain unsorted", func() {
+				So(*groups.Groups[0].Description, ShouldEqual, "B Group")
+				So(*groups.Groups[1].Description, ShouldEqual, "C Group")
+				So(*groups.Groups[2].Description, ShouldEqual, "A Group")
+			})
+		})
```
lindenmckenzie: would add a test for case insensitivity as well

features/steps/steps.go
```diff
@@ -33,6 +33,7 @@ func (c *IdentityComponent) RegisterSteps(ctx *godog.ScenarioContext) {
 	ctx.Step(`^an internal server error is returned from Cognito$`, c.anInternalServerErrorIsReturnedFromCognito)
 	ctx.Step(`^group "([^"]*)" and description "([^"]*)" exists in the database$`, c.groupAndDescriptionExistsInTheDatabase)
 	ctx.Step(`^group "([^"]*)" exists in the database$`, c.groupExistsInTheDatabase)
+	ctx.Step(`^group "([^"]*)" exists in the database to sort$`, c.groupExistsInTheDatabaseSort)
```
lindenmckenzie: do we need this differentiation with line 35?

api/groups.go
```diff
@@ -601,3 +612,49 @@ func (api *API) GetTeamsReportLines(listOfGroups *cognitoidentityprovider.ListGr
 
 	return &GroupsUsersList, nil
 }
+
+func sortGroups(ctx context.Context, listGroupOutput *cognitoidentityprovider.ListGroupsOutput, sortBy []string) {
```
lindenmckenzie: missing godoc.

lindenmckenzie (commented):
few comments.

api/groups.go
lindenmckenzie: this is the case for a few.

features/steps/steps.go
andre-urbani: > do we need this differentiation with line 35?

Kind of hard to explain without a demonstration, but the original `groupExistsInTheDatabase` function doesn't use the group value we need for sorting. It just creates a `Group`, whereas we need `*cognitoidentityprovider.ListGroupsOutput`. They both exist within the same struct but are otherwise unrelated, so I couldn't figure out any way to get around this other than creating a whole new set of functions and a new step. The existing functions couldn't really be modified from what I could see

api/groups.go
```diff
@@ -601,3 +612,52 @@ func (api *API) GetTeamsReportLines(listOfGroups *cognitoidentityprovider.ListGr
...
+		case "desc":
+			sortByGroupName(groups, false)
+			return
+		default:
+			http.Error(w, "incorrect sort value. Groups not sorted.", http.StatusBadRequest)
+			return
+		}
+	default:
+		http.Error(w, "incorrect sort value. Groups not sorted.", http.StatusBadRequest)
+		return
```
lindenmckenzie: This is going to keep processing rather than stopping the request I think? Should return an err and handle it above.

features/steps/steps.go
```diff
@@ -33,6 +33,7 @@ func (c *IdentityComponent) RegisterSteps(ctx *godog.ScenarioContext) {
 	ctx.Step(`^an internal server error is returned from Cognito$`, c.anInternalServerErrorIsReturnedFromCognito)
 	ctx.Step(`^group "([^"]*)" and description "([^"]*)" exists in the database$`, c.groupAndDescriptionExistsInTheDatabase)
 	ctx.Step(`^group "([^"]*)" exists in the database$`, c.groupExistsInTheDatabase)
+	ctx.Step(`^list group "([^"]*)" exists in the database$`, c.listGroupExistsInTheDatabase)
```
lindenmckenzie:
```suggestion
	ctx.Step(`^group "([^"]*)" exists in a list in the database$`, c.listGroupExistsInTheDatabase)
```

maybe?

cognito/mock/list_group_mocker.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func (m *CognitoIdentityProviderClientStub) AddListGroupWithName(name string) error {
+	newGroup, err := m.GenerateListGroup(name)
+	if err != nil {
+		return err
+	}
+	m.GroupsList = append(m.GroupsList, newGroup)
+	return nil
+}
```
lindenmckenzie, line 81: There was some other functions in that other file too, could you move them over?

lindenmckenzie (commented):
Some things to think about.

features/groups.feature
```diff
@@ -859,6 +859,178 @@ Feature: Groups
...
+                }
+            """
+
+    Scenario: GET /v1/groups?sort=created and checking the response status 200
+        Given group "B Group" exists in a list in the database
+        And group "A Group" exists in a list in the database
+        And group "C Group" exists in a list in the database
+        And I am an admin user
+        When I GET "/v1/groups?sort=created"
+        Then I should receive the following JSON response with status "200":
```
lindenmckenzie, line 997: Where's our component test for 400?

swagger.yaml
```diff
@@ -402,6 +402,17 @@ paths:
...
+        - in: query
+          name: sort
+          type: string
+          required: false
+          default: created
+          enum:
+            - created
+            - name
+            - name:asc
+            - name:desc
```
lindenmckenzie, line 415: needs a 400 response on here too.

lindenmckenzie (changes requested):
Couple of bits to update with the 400. Also - have a look at adding this to the sdk.

lindenmckenzie (approved):

PR #204, opened 2024-04-24
Implement sort users by forename asc and desc

dionisislolas:
### What

Add sort parameter to /groups/{id}/members endpoint

### How to review

Checkout this branch
Run the auth stack and navigate to http://localhost:25600/v1/groups/5/members?sortBy=forename:asc
You should see the users in ascending order
Navigate to http://localhost:25600/v1/groups/5/members?sortBy=forename:desc
You should see the users in descending order

### Who can review

Anyone

### DIFF
--- a/api/groups.go
+++ b/api/groups.go
@@ -6,12 +6,14 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"github.com/ONSdigital/dp-identity-api/models"
+	dplogs "github.com/ONSdigital/log.go/v2/log"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
 	"io"
 	"net/http"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


api/groups.go
```diff
@@ -185,13 +187,46 @@ func (api *API) ListUsersInGroupHandler(ctx context.Context, w http.ResponseWrit
 	listOfUsers := models.UsersList{}
 	listOfUsers.MapCognitoUsers(&listUsers)
 
+	if err = req.ParseForm(); err != nil {
+		dplogs.Error(ctx, "error parsing form", err)
```
gedge, line 191: should this return an error?

api/groups_test.go
```diff
@@ -2701,6 +2701,78 @@ func TestListGroupsUsersHandler(t *testing.T) {
 	})
 }
 
+func TestSortUsers(t *testing.T) {
+	Convey("Given we have some test users", t, func() {
+		json1 := "{\n  \"count\": 9,\n  \"users\": [\n    {\n      \"forename\": \"DTestForename\",\n      \"lastname\": \"LTestSurname\",\n      \"email\": \"DTestForename.LTestSurname@ons.gov.uk\",\n      \"groups\": [],\n      \"status\": \"CONFIRMED\",\n      \"active\": true,\n      \"id\": \"1234\",\n      \"status_notes\": \"\"\n    },\n    {\n      \"forename\": \"ATestForename\",\n      \"lastname\": \"HTestSurname\",\n      \"email\": \"ATestForename.HTestSurname@ons.gov.uk\",\n      \"groups\": [],\n      \"status\": \"CONFIRMED\",\n      \"active\": true,\n      \"id\": \"1234\",\n      \"status_notes\": \"\"\n    },\n    {\n      \"forename\": \"OTestForename\",\n      \"lastname\": \"STestSurname\",\n      \"email\": \"OTestForename.STestSurname@ons.gov.uk\",\n      \"groups\": [],\n      \"status\": \"CONFIRMED\",\n      \"active\": true,\n      \"id\": \"1234\",\n      \"status_notes\": \"\"\n    } ],\n  \"PaginationToken\": \"\"\n}"
```
gedge: use of backticks here to quote a multiline string might be easier to read

swagger.yaml
```diff
@@ -703,6 +703,60 @@ paths:
           description: "Unexpected Cognito response"
           schema:
             $ref: '#/definitions/ErrorResponse'
+  /groups/{id}/members?sortBy=forename:asc:
```
gedge: I don't see this query as another endpoint as this entry suggests (is this copied from another swagger spec?)

I see the query params acting more like `limit` [here](https://github.com/ONSdigital/dp-dataset-api/blob/develop/swagger.yaml#L238) as a parameter to an existing endpoint

swagger.yaml
```diff
@@ -690,6 +690,12 @@ paths:
           type: string
           required: true
           description: the group's ID
+        - in: query
+          name: sort
+          description: The sort order of the returned set of users
```
gedge, line 695: an example would be ideal 🙏

gedge: Hmm, I don't like that comment (is it rendered?).

Given that [examples can only be singular](https://swagger.io/docs/specification/2-0/adding-examples/) then I suggest using an [enum](https://swagger.io/docs/specification/2-0/enums/) for the small number of options for this param

gedge: e.g.
```yaml
   enum: [ "forename:asc", "forename:desc" ]
```
or whatever it is

gedge (approved):
LGTM

PR #206, opened 2024-04-29
Feature/add sort to users

AWitcherONS:
### What

User Requst to be able to sort users 

### How to review

Code should conform to current standards 
Unit tests 
Component tests
and Swagger

### Who can review

Anyone but me

### DIFF
--- a/api/users_test.go
+++ b/api/users_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"time"
 
 	"net/http"
@@ -24,6 +25,9 @@ const usersEndPoint = "http://localhost:25600/v1/users"
 const usersEndPointWithActiveFilterTrue = "http://localhost:25600/v1/users?active=true"
 const usersEndPointWithActiveFilterFalse = "http://localhost:25600/v1/users?active=false"
 const usersEndPointWithActiveFilterError = "http://localhost:25600/v1/user?active=false"
+
+const usersEndPointWithSortByEmail = "http://localhost:25600/v1/users?sort=email"
+const usersEndPointWithSortByEmailAcs = "http://localhost:25600/v1/users?sort=email:asc"
 
 const userEndPoint = "http://localhost:25600/v1/users/abcd1234"
 const changePasswordEndPoint = "http://localhost:25600/v1/users/self/password"
@@ -363,6 +367,75 @@ func TestListUserHandlerWithFilter(t *testing.T) {
 					So(errorResponse.Status, ShouldEqual, 400)
 					So(errorResponse.Errors[0].Error(), ShouldResemble, "InvalidFilterQuery")
 
+				},
+			},
+		}
+
+		for _, tt := range listUsersTest {
+			Convey(tt.description, func() {
+				m.ListUsersFunc = tt.listUsersFunction
+				r := tt.endpoint
+				successResponse, errorResponse := api.ListUsersHandler(ctx, w, r)
+				tt.assertions(successResponse, errorResponse)
+			},
+			)
+		}
+	})
+}
+
+func TestListUserHandlerWithSort(t *testing.T) {
+	var ctx = context.Background()
+	api, w, m := apiSetup()
+
+	Convey("List user - check expected responses", t, func() {
+		listUsersTest := []struct {
+			description       string
+			endpoint          *http.Request
+			listUsersFunction func(userInput *cognitoidentityprovider.ListUsersInput) (*cognitoidentityprovider.ListUsersOutput, error)
+			assertions        func(successResponse *models.SuccessResponse, errorResponse *models.ErrorResponse)
+		}{
+			{
+				"200 response from Cognito ",
+				httptest.NewRequest(http.MethodGet, usersEndPointWithSortByEmail, nil),
+				func(userInput *cognitoidentityprovider.ListUsersInput) (*cognitoidentityprovider.ListUsersOutput, error) {
+
+					var cognitoUsersList = []*cognitoidentityprovider.UserType{}
+					var forename, surname, email, status, id string = "Bob", "Smith", "email@ons.gov.uk", "CONFIRMED", "user-1"
+					cognitoUser := cognitoidentityprovider.UserType{
+						Attributes: []*cognitoidentityprovider.AttributeType{
+							{
+								Name:  aws.String("given_name"),
+								Value: &forename,
+							},
+							{
+								Name:  aws.String("family_name"),
+								Value: &surname,
+							},
+							{
+								Name:  aws.String("email"),
+								Value: &email,
+							},
+						},
+						UserStatus: &status,
+						Username:   &id,
+						Enabled:    aws.Bool(true),
+					}
+
+					cognitoUsersList = append(cognitoUsersList, &cognitoUser)
+
+					users := &cognitoidentityprovider.ListUsersOutput{
+						Users: cognitoUsersList,
+					}
+					return users, nil
+				},
+				func(successResponse *models.SuccessResponse, errorResponse *models.ErrorResponse) {
+					So(errorResponse, ShouldBeNil)
+					So(successResponse, ShouldNotBeNil)
+					So(successResponse.Status, ShouldEqual, 200)
+
+					var responseBody = cognitoidentityprovider.ListUsersOutput{}
+					json.Unmarshal(successResponse.Body, &responseBody)
+					fmt.Println(responseBody.Users)
 				},
 			},
 		}
--- a/api/users_test.go
+++ b/api/users_test.go
@@ -29,7 +29,7 @@ const usersEndPointWithSortByEmail = "http://localhost:25600/v1/users?sort=email
 const usersEndPointWithSortByEmailAsc = "http://localhost:25600/v1/users?sort=email:asc"
 const usersEndPointWithSortByEmailDesc = "http://localhost:25600/v1/users?sort=email:desc"
 const usersEndPointWithSortBy2fieldsDesc = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc"
-const usersEndPointWithSortBy2knowFieldsandUnknown = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc,dog"
+const usersEndPointWithSortBy2KnownFieldsAndUnknown = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc,dog"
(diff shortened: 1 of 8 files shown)


api/users_test.go
```diff
@@ -25,6 +25,9 @@ const usersEndPointWithActiveFilterTrue = "http://localhost:25600/v1/users?activ
 const usersEndPointWithActiveFilterFalse = "http://localhost:25600/v1/users?active=false"
 const usersEndPointWithActiveFilterError = "http://localhost:25600/v1/user?active=false"
 
+const usersEndPointWithSortByEmail = "http://localhost:25600/v1/users?sort=email"
+const usersEndPointWithSortByEmailAcs = "http://localhost:25600/v1/users?sort=email:asc"
```
lindenmckenzie:
```suggestion
const usersEndPointWithSortByEmailAsc = "http://localhost:25600/v1/users?sort=email:asc"
```

features/steps/steps.go
```diff
@@ -279,3 +287,24 @@ func (c *IdentityComponent) theResponseHeaderShouldContain(key, value string) (e
...
+	}
+	fmt.Println("\n\t--- actual ----", resBody)
+	fmt.Println("\n\t--- expected ----", expected)
+
+	if err = json.Unmarshal(resBody, &actual); err != nil {
+		return err
+	}
+
+	return nil
+}
```
lindenmckenzie: do we actually need a separate function here? Could we just use the same step description and point to the normal 'receive the following json' response?

features/users.feature
```diff
@@ -269,6 +269,334 @@ Feature: Users
...
+        And a user with forename "Mary", lastname "Martin", email "email7@ons.gov.uk", id "id_3" and password "Passw0rd!" exists in the database
+        And I am an admin user
+        When I GET "/v1/users?sort=id"
+        Then I should receive the following JSON response with status "200":
+            """
+            {
+              "PaginationToken": "",
+              "count": 3,
+              "users": [
+                              {
```
lindenmckenzie: indentation.

features/steps/steps.go
lindenmckenzie: Looking at the feature file, I don't think you're actually using this.

swagger.yaml
```diff
@@ -147,6 +147,10 @@ paths:
           name: active
           type: boolean
           description: filter on user active state.
+        - in: query
+          name: sort
+          type: string
+          description: sort the response by direction given key:asc or key:desc default asc, comma separates multi-field sorts
```
lindenmckenzie: could we note the default here as well please?

features/users.feature
```diff
@@ -269,6 +269,334 @@ Feature: Users
...
+            }
+            """
+
+    @get-users-list
+    Scenario: GET /v1/users and checking the response status 200 with sort by id:desc
+        Given a user with forename "Adam", lastname "Adams", email "email5@ons.gov.uk", id "id_2" and password "Passw0rd!" exists in the database
+        And a user with forename "William", lastname "Williams", email "email9@ons.gov.uk", id "id_1" and password "Passw0rd!" exists in the database
+        And a user with forename "Mary", lastname "Martin", email "email7@ons.gov.uk", id "id_3" and password "Passw0rd!" exists in the database
+        And I am an admin user
+        When I GET "/v1/users?sort=id:asc"
```
lindenmckenzie: Think this should be `:desc` from the scenario description. 

Which means the json needs to be inverted.

features/users.feature
```diff
@@ -269,6 +269,334 @@ Feature: Users
...
+                  "forename": "William",
+                  "groups": [],
+                  "id": "id_5",
+                  "lastname": "Adams",
+                  "status": "CONFIRMED",
+                  "status_notes": ""
+                }
+                ]
+            }
+            """
```
lindenmckenzie, line 611: what happens if someone puts in:

`/users?sort=dogs:asc`

query/sortUser.go
```diff
@@ -0,0 +1,103 @@
...
+		case less(p, q):
+			return true
+		case less(q, p):
+			return false
+		}
+	}
+	return ms.less[k](p, q)
+}
+
+func SortBy(paramsSlice string, arr []models.UserParams) {
```
lindenmckenzie: I'm not sure about how descriptive the 'paramsSlice' is. 

Godoc would be useful here too (and on the above docs)

features/steps/steps.go
```diff
@@ -46,6 +47,8 @@ func (c *IdentityComponent) RegisterSteps(ctx *godog.ScenarioContext) {
 	ctx.Step(`^request header Accept is "([^"]*)"$`, c.requestHeaderAcceptIs)
 	ctx.Step(`^the response should match the following csv:$`, c.theResponseShouldMatchTheFollowingCsv)
 	ctx.Step(`^the response header "([^"]*)" should contain "([^"]*)"$`, c.theResponseHeaderShouldContain)
+	ctx.Step(`^I should receive the following sorted JSON response with status "([^"]*)":$`, c.iShouldReceiveTheFollowingSortedJSONResponseWithStatus)
```
lindenmckenzie: Not sure this is being used in the feature file.

query/sortUser.go
```diff
@@ -0,0 +1,103 @@
...
+	emailDesc := func(c1, c2 *models.UserParams) bool { return c1.Email > c2.Email }
+	idAsc := func(c1, c2 *models.UserParams) bool { return c1.ID < c2.ID }
+	idDesc := func(c1, c2 *models.UserParams) bool { return c1.ID > c2.ID }
+
+	p1 := strings.Split(paramsSlice, ",")
+
+	var orderFunc []lessFunc
+	for _, p2 := range p1 {
+		p3 := strings.Split(p2, ":")
+		if len(p3) > 1 {
```
lindenmckenzie: I think these `px` variables could do with some more descriptive names.

query/sortUser.go
```diff
@@ -0,0 +1,103 @@
...
+
+func SortBy(paramsSlice string, arr []models.UserParams) {
+	forenameAsc := func(c1, c2 *models.UserParams) bool { return c1.Forename < c2.Forename }
+	forenameDesc := func(c1, c2 *models.UserParams) bool { return c1.Forename > c2.Forename }
+	lastnameAsc := func(c1, c2 *models.UserParams) bool { return c1.Lastname < c2.Lastname }
+	lastnameDesc := func(c1, c2 *models.UserParams) bool { return c1.Lastname > c2.Lastname }
+	emailAsc := func(c1, c2 *models.UserParams) bool { return c1.Email < c2.Email }
+	emailDesc := func(c1, c2 *models.UserParams) bool { return c1.Email > c2.Email }
+	idAsc := func(c1, c2 *models.UserParams) bool { return c1.ID < c2.ID }
+	idDesc := func(c1, c2 *models.UserParams) bool { return c1.ID > c2.ID }
```
lindenmckenzie: Could this be made more generic by passing attributes in? i.e. we could map the incoming attributes to the underlying struct variables. Covers a bit of those nested if statements later as well.

swagger.yaml
lindenmckenzie: an enum of all the options would be good too.

lindenmckenzie (commented):
Some comments - happy to have a longer discussion on it tomorrow in the office if you're in :)

api/users_test.go
AWitcherONS: :asc is not mandatory however :desc is mandatory 
 I am testing both

features/users.feature
AWitcherONS: Good test ... it shouldn't fail just will not sort the query

api/users_test.go
```diff
@@ -24,7 +25,11 @@ const usersEndPoint = "http://localhost:25600/v1/users"
 const usersEndPointWithActiveFilterTrue = "http://localhost:25600/v1/users?active=true"
 const usersEndPointWithActiveFilterFalse = "http://localhost:25600/v1/users?active=false"
 const usersEndPointWithActiveFilterError = "http://localhost:25600/v1/user?active=false"
-
+const usersEndPointWithSortByEmail = "http://localhost:25600/v1/users?sort=email"
+const usersEndPointWithSortByEmailAsc = "http://localhost:25600/v1/users?sort=email:asc"
+const usersEndPointWithSortByEmailDesc = "http://localhost:25600/v1/users?sort=email:desc"
+const usersEndPointWithSortBy2fieldsDesc = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc"
+const usersEndPointWithSortBy2knowFieldsandUnknown = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc,dog"
```
lindenmckenzie:
```suggestion
const usersEndPointWithSortBy2KnownFieldsAndUnknown = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc,dog"
```

api/users_test.go
```diff
@@ -24,7 +25,11 @@ const usersEndPoint = "http://localhost:25600/v1/users"
 const usersEndPointWithActiveFilterTrue = "http://localhost:25600/v1/users?active=true"
 const usersEndPointWithActiveFilterFalse = "http://localhost:25600/v1/users?active=false"
 const usersEndPointWithActiveFilterError = "http://localhost:25600/v1/user?active=false"
-
+const usersEndPointWithSortByEmail = "http://localhost:25600/v1/users?sort=email"
+const usersEndPointWithSortByEmailAsc = "http://localhost:25600/v1/users?sort=email:asc"
+const usersEndPointWithSortByEmailDesc = "http://localhost:25600/v1/users?sort=email:desc"
+const usersEndPointWithSortBy2fieldsDesc = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc"
```
lindenmckenzie:
```suggestion
const usersEndPointWithSortBy2FieldsDesc = "http://localhost:25600/v1/users?sort=forename:desc,lastname:desc"
```

swagger.yaml
```diff
@@ -697,7 +718,6 @@ paths:
           required: false
           default: created
           enum:
-            - created
```
lindenmckenzie: this is being removed from get groups?

swagger.yaml
```diff
@@ -147,6 +147,27 @@ paths:
           name: active
           type: boolean
           description: filter on user active state.
+        - in: query
+          name: sort
+          type: string
```
lindenmckenzie, line 152:
```suggestion
          type: string
          default: id:asc
```

lindenmckenzie (commented):
Some trivial comments.

lindenmckenzie (approved):

PR #212, opened 2024-06-17
Fix/fix refresh expiry

lindenmckenzie:
### What

dp-identity-api was setting a refresh token expiry date of 12 days when in fact it is 12 hours. 

This is because the api anticipates cognito will send it a unit in days - I've accounted for this and now it sets for 12 hours. 

### How to review

Connect to sandbox cognito using the instructions in the readme. 
Hit http://localhost:25600/v1/tokens
Check the refreshTokenExpirationTime to ensure it's now correct.

### Who can review

Backend dev.


AWitcherONS (dismissed):
Is there any Documentation for the new Makefile option of debug-watch
What does this do and why

AWitcherONS (approved):

PR #229, opened 2024-11-20
Feature/update spec

lindenmckenzie:
### What

Refactored swagger specification to reflect API better.
Added redocly for API validation. 

### How to review

Review spec against responses from API / internal clarity - get endpoints are most easily reviewed by going to sandbox florence, logging in and then hitting /api/v1/{endpoint} - not all endpoints are available here.

### Who can review

Not me.

### DIFF
--- a/ci/lint.yml
+++ b/ci/lint.yml
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.23.2-bullseye
+    tag: 1.23.1-bullseye-node-20
 
 inputs:
   - name: dp-identity-api
--- a/ci/lint.yml
+++ b/ci/lint.yml
@@ -5,7 +5,7 @@ platform: linux
 image_resource:
   type: docker-image
   source:
-    repository: golang
+    repository: onsdigital/dp-concourse-tools-node-go
     tag: 1.23.1-bullseye-node-20
 
 inputs:
--- a/redocly.yaml
+++ b/redocly.yaml
@@ -1,0 +1,4 @@
+extends:
+  - recommended
+rules:
+  operation-operationId: off
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -49,19 +49,15 @@ paths:
           schema:
             $ref: '#/definitions/SignInExpirationTime'
         400:
-          description: "Invalid request. Empty or invalid email or empty password."
-          schema:
-            $ref: '#/definitions/ErrorResponse'
+          $ref: '#/responses/BadRequestError'
         401:
           $ref: '#/responses/UnauthorizedError'
         403:
           description: "Forbidden. Too many login attempts"
           schema:
-            $ref: '#/definitions/ErrorResponse'
+            $ref: '#/definitions/ErrorList'
         500:
-          description: "Unexpected internal error"
-          schema:
-            $ref: '#/definitions/ErrorResponse'
+          $ref: '#/responses/InternalError'
     delete:
       security:
         - Authorization: []
@@ -76,9 +72,7 @@ paths:
         401:
           $ref: '#/responses/UnauthorizedError'
         500:
-          description: "Unexpected internal error"
-          schema:
-            $ref: '#/definitions/ErrorResponse'
+          $ref: '#/responses/InternalError'
   /tokens/self:
     put:
       tags:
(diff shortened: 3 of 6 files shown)


ci/lint.yml
```diff
@@ -5,8 +5,8 @@ platform: linux
 image_resource:
   type: docker-image
   source:
-    repository: golang
-    tag: 1.23.2-bullseye
+    repository: onsdigital/dp-concourse-tools-node-go
+    tag: 1.23.1-bullseye-node-20
```
janderson2, line 9: I wonder if this could be a use case for a [custom megalinter plugin](https://megalinter.io/7.0.1/plugins/#create-your-own-plugin)? That way we could avoid the need for these combined images. Maybe something to consider as a future iteration?

redocly.yaml
```diff
@@ -0,0 +1,4 @@
+extends:
+  - recommended
+rules:
+  operation-operationId: off
```
janderson2, line 4: Had to look this field up. Agree with this being disabled as it is optional and not required for our use cases.

swagger.yaml
```diff
@@ -805,6 +898,8 @@ paths:
 responses:
   InternalError:
     description: "Failed to process the request due to an internal error"
+  UnauthorizedError:
+    description: Authentication information is missing or invalid
```
janderson2: Should these responses have a response body (i.e. schema) or do they have an empty body?

swagger.yaml
```diff
@@ -316,6 +321,39 @@ paths:
...
+        401:
+          $ref: '#/responses/UnauthorizedError'
+        404:
+          description: "User not found"
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        500:
+          description: "Internal error occurred"
+          schema:
+            $ref: '#/definitions/ErrorResponse'
```
janderson2: There appear to be a number of repeated responses such as this 500 response with the same model. There is also a global `InternalError` response defined. Consider updating the spec to use this globally defined response.

This also applies to the 401 responses and could be applied to the 400s if the descriptions were made to be more generic.

swagger.yaml
```diff
@@ -778,16 +811,74 @@ paths:
...
+      description: "To return a list of :-
+          group description
+          group member user email
+          json by default 
+          or if Request Header Accept 'text/csv' then csv format, consisting of preview team description amd user email. 
+        Notable Consideration :- 
+          Included on report
+            groups acting as a role (i.e. publishers, admin)
+          Excluded from Report
+            groups that are empty of users"
```
janderson2: This description was hard to understand and the formatting appears to be off. Consider enabling rich formatting for this field using `|` and potentially clarifying the wording:
```suggestion
      description: |
        The list of group memebership with the following fields:
    
        * group description
        * group member user email
    
        Format is JSON by default or csv format if Request Header Accept 'text/csv' is specified.
        
        Included on report: groups acting as a role (i.e. publishers, admin)
        Excluded from Report: groups that are empty of users"
```

swagger.yaml
```diff
@@ -778,16 +811,74 @@ paths:
...
+        - in: header
+          name: Accept:application/json
+          type: string
+          required: false
+          description: "Groups report as JSON"
+        - in: header
+          name: Accept:text/csv
+          type: string
+          required:  false
+          description: "Groups report as CSV"
```
janderson2: I believe it would be better practice to combine these and use an enum:
```suggestion
        - in: header
          name: Accept
          type: string
          enum:
            - "application/json"
            - "text/csv"
          required: false
          description: "Whether a JSON or CSV report is requested"
```

janderson2 (commented):
Really like the linting 🙌  Have added a couple of comments on the spec.

ci/lint.yml
lindenmckenzie: Yeah it absolutely is, but until we've done a bit more with it I didn't want to integrate quite yet for this spec refactor

swagger.yaml
lindenmckenzie: Yes, agree, was just trying to draw the line somewhere!

swagger.yaml
lindenmckenzie: Agree

swagger.yaml
lindenmckenzie: it's an empty body

janderson2: That is probably a defect we should log somewhere as all errors should return an error message body.

janderson2 (approved):
:heart:

PR #233, opened 2025-04-07
Upgrade aws sdk go to v2

cookel2:
### What

The aws sdk for go has been upgraded to:

github.com/aws/aws-sdk-go-v2 v1.36.3

Consequently the version of go has been upgraded to 1.24, and the following other libraries have also either been added or upgraded to:

github.com/ONSdigital/dp-api-clients-go/v2 v2.264.0
github.com/ONSdigital/dp-authorisation/v2 v2.32.0
github.com/ONSdigital/dp-component-test v0.17.0
github.com/ONSdigital/dp-net/v3 v3.1.0
github.com/ONSdigital/dp-permissions-api v0.27.0
github.com/ONSdigital/log.go/v2 v2.4.4
github.com/aws/aws-sdk-go-v2/config v1.29.12
github.com/aws/aws-sdk-go-v2/service/cognitoidentityprovider v1.51.3
github.com/aws/aws-sdk-go-v2/service/s3 v1.78.2
github.com/aws/smithy-go v1.22.3
github.com/golang-jwt/jwt/v4 v4.5.2
github.com/stretchr/testify v1.10.0

Changes have been made to fix any resulting errors in the lint, unit, or component tests.

For compatibility with go 1.24, the linter version has been upgraded to:

github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8

### How to review

Check that the changes look correct and all the tests pass. You can also make sure that it runs successfully in the Auth stack.

### Who can review

Anyone but me!


Dockerfile.local
```diff
@@ -1,4 +1,6 @@
-FROM golang:1.23.4-bullseye AS base
+FROM golang:1.24.1-bullseye AS base
```
Oluwawunmi: 1.24.2 is now available

ci/component.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.23.4-bullseye
+    tag: 1.24.1-bullseye
```
Oluwawunmi: 1.24.2-bullseye is now available

ci/lint.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: onsdigital/dp-concourse-tools-node-go
-    tag: 1.23.4-bullseye-node-20
+    tag: 1.24.1-bullseye-node-20
```
Oluwawunmi: 1.24.2-bullseye-node-20 is now available

ci/unit.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.23.4-bullseye
+    tag: 1.24.1-bullseye
```
Oluwawunmi: 1.24.2-bullseye is now available

.golangci.yml
```diff
@@ -59,7 +59,6 @@ linters:
     - gofmt
     - goimports
     - revive
-    - gosec
```
Oluwawunmi: Instead of removing it completely, why not just exclude it on the mocker files or ignore the individual lines

models/users_test.go
```diff
@@ -433,7 +439,7 @@ func TestUserParams_BuildSuccessfulJsonResponse(t *testing.T) {
 		err = json.Unmarshal(response, &userJSON)
 		So(err, ShouldBeNil)
 		So(userJSON["id"], ShouldEqual, name)
-		So(userJSON["status"], ShouldEqual, status)
+		So(userJSON["status"], ShouldEqual, "UNCONFIRMED")
```
Oluwawunmi: Maybe make this the "types.UserStatusTypeUnconfirmed" instead of the string

Oluwawunmi: And for other instances

api/tokens.go
```diff
@@ -18,32 +20,33 @@ func (api *API) TokensHandler(ctx context.Context, _ http.ResponseWriter, req *h
 			_ = models.NewError(ctx, err, models.BodyCloseError, models.BodyClosedFailedDescription)
 		}
 	}()
-
+	println("in TokensHandler 1")
```
Oluwawunmi: Rather log.info instead of printLn.

models/users.go
```diff
@@ -270,20 +277,24 @@ func (p UserParams) MapCognitoDetails(userDetails *cognitoidentityprovider.UserT
 			statusNotes = *attr.Value
 		}
 	}
+	println("The ID will be: " + *userDetails.Username)
```
Oluwawunmi: Lets. log if necessary or remove

api/users.go
```diff
@@ -135,7 +137,14 @@ func (api *API) GetUserHandler(ctx context.Context, _ http.ResponseWriter, req *
 		return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, responseErr)
 	}
 
-	user.MapCognitoGetResponse(userResp)
+	userDetails := user.MapCognitoGetResponse(userResp)
+	user.Active = userDetails.Active
```
Oluwawunmi: not sure why we map to userDetails first and then essentially mapping userDetails to user immediately afterwards. Maybe a method to do it once and for all?

ci/build.yml
```diff
@@ -6,7 +6,7 @@ image_resource:
   type: docker-image
   source:
     repository: golang
-    tag: 1.23.4-bullseye
+    tag: 1.24.1-bullseye
```
Oluwawunmi: 1.24.2-bullseye is now available

Oluwawunmi (approved):
Just a few minor adjustments

Oluwawunmi (changes requested):
Just a few minor adjustments

.golangci.yml
cookel2: I looked into excluding individual lines previously but I think there were too many. I did think it would be good to find an alternative instead of removing it completely though. How do I exclude it on the mocker files?

ci/component.yml
cookel2: 👍

Dockerfile.local
cookel2: 👍

ci/lint.yml
cookel2: 👍

models/users_test.go
cookel2: I just tried that but the test failed. However, I've added "UNCONFIRMED" as a constant instead.

models/users.go
cookel2: 👍

.golangci.yml
lindenmckenzie: Agree - there's more nuanced ways to achieve this. @Oluwawunmi can take you through it I'm sure.

Oluwawunmi: when I ran locally, it was just two files (mocker). But you can exclude it similar to the way revive was excludes on the test.go files in this same file.

.golangci.yml
```diff
@@ -80,5 +79,6 @@ issues:
       text: "dot-imports"
       linters:
         - revive
+        - gosec
       source: "github.com/smartystreets/goconvey/convey"
```
lindenmckenzie: I don't think this is doing what you think - specifically this rule is stopping the 'dot-imports' rule from the 'revive' linter. Adding gosec here won't do anything, unless it also has a 'dot-imports' rule.

api/api_test.go
```diff
@@ -165,21 +169,22 @@ func hasRoute(r *mux.Router, path, method string) bool {
...
 func apiSetup() (*API, *httptest.ResponseRecorder, *mock.MockCognitoIdentityProviderClient) {
 	var (
-		ctx                                                 = context.Background()
-		r                                                   = mux.NewRouter()
-		poolID, clientID, clientSecret, awsRegion, authFlow = "us-west-11_bxushuds", "client-aaa-bbb", "secret-ccc-ddd", "eu-west-1234", "USER_PASSWORD_AUTH"
-		allowedDomains                                      = []string{"@ons.gov.uk", "@ext.ons.gov.uk"}
+		ctx                                       = context.Background()
+		r                                         = mux.NewRouter()
+		poolID, clientID, clientSecret, awsRegion = "us-west-11_bxushuds", "client-aaa-bbb", "secret-ccc-ddd", "eu-west-1234"
+		authFlow                                  = types.AuthFlowTypeUserPasswordAuth // "USER_PASSWORD_AUTH"
```
lindenmckenzie:
```suggestion
		authFlow                                  = types.AuthFlowTypeUserPasswordAuth
```
Don't think comment is warranted.

api/groups.go
```diff
@@ -175,9 +177,9 @@ func (api *API) ListUsersInGroupHandler(ctx context.Context, _ http.ResponseWrit
 	vars := mux.Vars(req)
 	group := models.Group{ID: vars["id"]}
 
-	listOfUsersInput := []*cognitoidentityprovider.UserType{}
+	var listOfUsersInput []types.UserType
```
lindenmckenzie: I'm surprised this is no longer a pointer.

api/tokens.go
Oluwawunmi: Still some printLn statements in this file

api/tokens.go
```diff
@@ -18,32 +20,32 @@ func (api *API) TokensHandler(ctx context.Context, _ http.ResponseWriter, req *h
...
+	println("in TokensHandler 4")
 	if authErr != nil {
 		responseErr := models.NewCognitoError(ctx, authErr, "Cognito InitiateAuth request from sign in handler")
 		if responseErr.Code == models.InternalError {
 			return nil, models.NewErrorResponse(http.StatusInternalServerError, nil, responseErr)
 		}
-
+		println("The authErr description is: " + responseErr.Description)
 		switch responseErr.Description {
 		case models.SignInFailedDescription:
```
lindenmckenzie: These all look to be debug statements - could these be:
- removed
- logs
- moved to returned errors.

api/tokens.go
```diff
@@ -60,9 +62,9 @@ func (api *API) TokensHandler(ctx context.Context, _ http.ResponseWriter, req *h
 			return nil, models.NewErrorResponse(http.StatusBadRequest, nil, responseErr)
 		}
 	}
-
+	println("in TokensHandler 5")
```
lindenmckenzie: debug statement, see above.

api/users.go
cookel2: Good spot Wunmi. I had added those lines thinking that the user's ID value, which is set earlier in the method, might have been overwritten if I just set the user to be the return value. Anyway, I was wrong about that so I've deleted those unnecessary lines now.

api/tokens.go
cookel2: 👍

api/tokens_test.go
```diff
@@ -98,10 +100,13 @@ func TestAPI_TokensHandler(t *testing.T) {
 	})
 
 	Convey("Sign In Cognito internal error: adds an error to the ErrorResponse and sets its Status to 500", t, func() {
-		awsOrigErr := errors.New(awsErrCode)
-		awsErr := awserr.New(awsErrCode, awsErrMessage, awsOrigErr)
-		// mock failed call to: InitiateAuth(input *cognitoidentityprovider.InitiateAuthInput) (*cognitoidentityprovider.InitiateAuthOutput, error)
-		m.InitiateAuthFunc = func(_ *cognitoidentityprovider.InitiateAuthInput) (*cognitoidentityprovider.InitiateAuthOutput, error) {
+		awsOrigErr := smithy.ErrorFault(1) // server error
```
lindenmckenzie: I like these comments (as elsewhere) - but might be easier setting this to a const like

```
const serverError = smithy.ErrorFault(1)
```
Then it would be slightly cleaner.

ci/build.yml
cookel2: 👍

dummy-data/import-dummy-users/populate_dummy_data.go
```diff
@@ -132,17 +135,18 @@ func confirmUsers(ctx context.Context, client cognito.Client, userPoolID string,
 		}
 		for _, backoff := range backoffSchedule {
 			user := models.UserParams{}
-			passwordError := user.GeneratePassword(ctx)
+			var passwordError error
+			user.Password, passwordError = user.GeneratePassword(ctx)
```
lindenmckenzie: Why the move away from having the method on the user set it?

models/users.go
```diff
@@ -70,22 +72,25 @@ func (p UsersList) BuildListUserRequest(filterString, requiredAttribute string,
 }
 
 // MapCognitoUsers maps the users from the cognito response into the UsersList Users attribute and sets the Count attribute
-func (p *UsersList) MapCognitoUsers(cognitoResults *[]*cognitoidentityprovider.UserType) {
+func (p UsersList) MapCognitoUsers(cognitoResults *[]types.UserType) (users []UserParams, count int) {
 	p.Users = []UserParams{}
 	for _, user := range *cognitoResults {
 		p.Users = append(p.Users, UserParams{}.MapCognitoDetails(user))
 	}
+	println("The length of UserParams, in the UsersList, is: " + strconv.Itoa(len(p.Users)))
```
lindenmckenzie: This is a debug statement - see above.

models/users.go
```diff
@@ -70,22 +72,25 @@ func (p UsersList) BuildListUserRequest(filterString, requiredAttribute string,
...
 	p.Count = len(p.Users)
+	return p.Users, p.Count
 }
 
 // SetUsers sets the UsersList Users attribute and sets the Count attribute
-func (p *UsersList) SetUsers(usersList *[]UserParams) {
+func (p UsersList) SetUsers(usersList *[]UserParams) (users []UserParams, count int) {
```
lindenmckenzie: Why the move away from a pointer?

scripts/import_users/users/restore_users.go
```diff
@@ -53,29 +54,31 @@ func ImportUsersFromS3(ctx context.Context, cfg *config.Config) error {
...
 		log.Error(ctx, "failed to generate password", err)
 	}
 
 	createUserRequest := userInfo.BuildCreateUserRequest(userInfo.ID, cfg.AWSCognitoUserPoolID)
-	_, err := client.AdminCreateUser(createUserRequest)
-	if err != nil {
-		log.Error(ctx, fmt.Sprintf("failed to processline %v user: %+v", lineNumber, userInfo), err)
+	_, errCreateUser := client.AdminCreateUser(ctx, createUserRequest)
+	if errCreateUser != nil {
+		log.Error(ctx, fmt.Sprintf("failed to process line %v user: %+v", lineNumber, userInfo), err)
```
lindenmckenzie: better to log a generic error and then include the variables in a data block.

service/initialise.go
```diff
@@ -88,10 +89,17 @@ func (e *Init) DoGetHealthCheck(cfg *config.Config, buildTime, gitCommit, versio
 }
 
 // DoGetCognitoClient creates a CognitoClient with the provided region
-func (e *Init) DoGetCognitoClient(awsRegion string) cognitoclient.Client {
-	client := cognito.New(session.Must(session.NewSessionWithOptions(session.Options{
-		SharedConfigState: session.SharedConfigEnable,
-	})), &aws.Config{Region: &awsRegion})
+func (e *Init) DoGetCognitoClient(ctx context.Context, awsRegion string) cognitoclient.Client {
+	cfg, err := sdkcfg.LoadDefaultConfig(context.TODO(),
```
lindenmckenzie: Why `context.TODO()` here?

lindenmckenzie (changes requested):

lindenmckenzie:
I clicked submit to soon - I'm not sure on the reasoning behind some of the moving away from pointers in here without a much longer look at the code. Given the wide reaching nature of the change, I'd also like to do some testing in local environment with this.

.golangci.yml
cookel2: fixed now - thanks

Oluwawunmi (dismissed):
Pending the changes requested by Linden

api/groups.go
cookel2: It was difficult to tell whether it should be a pointer or not but I changed it to make tests pass.

dummy-data/import-dummy-users/populate_dummy_data.go
cookel2: It seemed to be the best way to fix the following error in models/users_test.go

  * /Users/cookel/EllasFiles/DpCodeRepos/dp-identity-api/models/users_test.go
  Line 146:
  Expected     ''
  to NOT equal ''
  (but it did)!

models/users.go
cookel2: I can't remember exactly but if I used a pointer for that function then it would give me errors if I didn't use a pointer for all the other functions in that class too. So it might have been one of the other functions that I had to remove the pointer from initially (and consequently I had to remove the pointer from all functions otherwise I would get errors about inconsistency).

models/users.go
cookel2: removed

scripts/import_users/users/restore_users.go
cookel2: done

service/initialise.go
cookel2: oops sorry! I think I must have copied it from an example. I've removed it now.

api/api_test.go
cookel2: 👍 - now deleted

api/tokens.go
cookel2: 👍 - done

api/tokens.go
cookel2: 👍 - deleted

api/tokens_test.go
cookel2: good idea - done

api/api_test.go
```diff
@@ -303,58 +308,70 @@ func TestInitialiseRoleGroups(t *testing.T) {
...
 			{
 				// admin group exists
-				func(input *cognitoidentityprovider.CreateGroupInput) (*cognitoidentityprovider.CreateGroupOutput, error) {
+				func(_ context.Context, input *cognitoidentityprovider.CreateGroupInput, _ ...func(*cognitoidentityprovider.Options)) (*cognitoidentityprovider.CreateGroupOutput, error) {
 					if *input.GroupName == models.AdminRoleGroup {
 						awsErrCode := "GroupExistsException"
 						awsErrMessage := "This group exists"
-						awsOrigErr := errors.New(awsErrCode)
-						awsErr := awserr.New(awsErrCode, awsErrMessage, awsOrigErr)
+						awsOrigErr := smithy.ErrorFault(2) // client error
```
lindenmckenzie: Same point on this file about generating the errors as with users_test or wherever it was updated before.

api/groups.go
lindenmckenzie: So I would question why we need to pass in the input at all? 

The function just appends to the list and returns it. Nowhere I can see does it actually send in a list which is not empty.

api/tokens.go
```diff
@@ -18,32 +20,28 @@ func (api *API) TokensHandler(ctx context.Context, _ http.ResponseWriter, req *h
 			_ = models.NewError(ctx, err, models.BodyCloseError, models.BodyClosedFailedDescription)
 		}
 	}()
-
 	body, err := io.ReadAll(req.Body)
 	if err != nil {
+		println("Returning a handleBodyReadError")
```
lindenmckenzie: Debug statement - please remove / turn into proper logging.

api/tokens.go
```diff
@@ -18,32 +20,28 @@ func (api *API) TokensHandler(ctx context.Context, _ http.ResponseWriter, req *h
...
 		return nil, handleBodyReadError(ctx, err)
 	}
-
 	var userSignIn models.UserSignIn
 	err = json.Unmarshal(body, &userSignIn)
 	if err != nil {
+		println("Returning a handleBodyUnmarshallError")
```
lindenmckenzie: Debug statement - please remove / turn into proper logging.

api/tokens_test.go
```diff
@@ -521,17 +539,20 @@ func TestSignOutAllUsersGoRoutine(t *testing.T) {
...
+				func(_ context.Context, signOutInput *cognitoidentityprovider.AdminUserGlobalSignOutInput, _ ...func(*cognitoidentityprovider.Options)) (*cognitoidentityprovider.AdminUserGlobalSignOutOutput, error) {
 					if *signOutInput.Username == userNamesList[3] {
 						awsErrCode := "TooManyRequestsException"
 						awsErrMessage := "Too many requets received"
-						awsOrigErr := errors.New(awsErrCode)
-						awsErr := awserr.New(awsErrCode, awsErrMessage, awsOrigErr)
+						awsErr := &smithy.GenericAPIError{
+							Code:    awsErrCode,
+							Message: awsErrMessage,
+							Fault:   smithy.ErrorFault(2), // client error
```
lindenmckenzie: Would abstract this in the same way as serverError.

api/users_test.go
```diff
@@ -251,24 +250,28 @@ func TestListUserHandler(t *testing.T) {
...
 				},
 				http.StatusOK,
 			},
 			{
 				// 500 response from Cognito
-				func(_ *cognitoidentityprovider.ListUsersInput) (*cognitoidentityprovider.ListUsersOutput, error) {
-					awsOrigErr := errors.New(awsErrCode)
-					awsErr := awserr.New(awsErrCode, awsErrMessage, awsOrigErr)
+				func(_ context.Context, _ *cognitoidentityprovider.ListUsersInput, _ ...func(*cognitoidentityprovider.Options)) (*cognitoidentityprovider.ListUsersOutput, error) {
+					awsOrigErr := smithy.ErrorFault(1) // server error
```
lindenmckenzie: Same here for serverError

scripts/import_users/groups/restore_groups.go
```diff
@@ -95,29 +95,33 @@ func ImportGroupsMembersFromS3(ctx context.Context, cfg *config.Config) error {
...
-func createGroup(ctx context.Context, lineNumber int, line []string, client *cognito.CognitoIdentityProvider, userPoolID string, colsMap map[string]int) {
+func createGroup(ctx context.Context, lineNumber int, line []string, client *cognito.Client, userPoolID string, colsMap map[string]int) {
 	createGroup := models.CreateUpdateGroup{}
 	createGroup.ID = &line[colsMap["groupname"]]
-	precedence, _ := strconv.Atoi(line[colsMap["precedence"]])
-	precedence1 := int64(precedence)
-	createGroup.Precedence = &precedence1
+	parsed, err := strconv.ParseInt(line[colsMap["precedence"]], 10, 32)
+	if err != nil {
+		log.Error(ctx, fmt.Sprintf("failed to parse precedence on line: %d (incl. header)", lineNumber+1), err)
```
lindenmckenzie: We prefer static log messages with a `log.Data` attached.

scripts/utils/aws_utils.go
```diff
@@ -2,47 +2,57 @@ package utils
...
-		SharedConfigState: session.SharedConfigEnable,
-	})))
+	cfg, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithRegion(region),
+		config.WithSharedConfigProfile(awsProfile),
+	)
+
+	if err != nil {
+		log.Fatal(ctx, "unable to load the SDK", err)
+	}
```
lindenmckenzie: Should anything else happen on error?

scripts/utils/aws_utils.go
```diff
@@ -2,47 +2,57 @@ package utils
...
-	})))
+func GetCognitoClient(ctx context.Context, awsProfile, region string) *cognitoidentityprovider.Client {
+	cfg, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithRegion(region),
+		config.WithSharedConfigProfile(awsProfile),
+	)
+
+	if err != nil {
+		log.Fatal(ctx, "unable to load the SDK", err)
+	}
```
lindenmckenzie: Should anything else happen on error?

service/initialise.go
```diff
@@ -88,10 +89,17 @@ func (e *Init) DoGetHealthCheck(cfg *config.Config, buildTime, gitCommit, versio
...
-		SharedConfigState: session.SharedConfigEnable,
-	})), &aws.Config{Region: &awsRegion})
+func (e *Init) DoGetCognitoClient(ctx context.Context, awsRegion string) cognitoclient.Client {
+	cfg, err := sdkcfg.LoadDefaultConfig(ctx,
+		sdkcfg.WithRegion(awsRegion),
+	)
+
+	if err != nil {
+		log.Fatal(ctx, "unable to load the SDK", err)
+	}
```
lindenmckenzie: Should anything else happen on error?

api/api_test.go
lindenmckenzie: N.B. there a few places this can be done - this is a general point about all the test files.

api/groups.go
cookel2: I agree that this functionality doesn't really seem to be used. The only place that I can find it passing in a user is in line 897 of api/groups_test.go where it is trying to convey "When GetUsersInAGroup in called with a list of users the appended list of users in returned". But that can be deleted.

api/api_test.go
cookel2: all done now (as far as I can tell) 👍

api/groups.go
cookel2: Refactored now 👍

api/tokens.go
cookel2: removed

scripts/import_users/groups/restore_groups.go
cookel2: 👍

api/tokens.go
cookel2: removed

scripts/utils/aws_utils.go
cookel2: I would say that it should exit/break but when I tried adding an exit statement after this line then the linter gave me this message: 
unreachable-code: unreachable code after this statement (revive)
		log.Fatal(ctx, "unable to load the SDK", err)

service/initialise.go
cookel2: Similarly if I add anything after this line then I get the following message: unreachable-code: unreachable code after this statement (revive)
		log.Fatal(ctx, "unable to load the SDK", err)

scripts/utils/aws_utils.go
cookel2: Please see comments below.

service/initialise.go
lindenmckenzie: So that's a false positive from the linter - our logging doesn't actually fail on log.Fatal.

Oluwawunmi (dismissed):
Approved pending Lindens review

service/initialise.go
cookel2: That's interesting, thanks, I've updated it now.

lindenmckenzie (approved):

PR #234, opened 2025-04-24
Revert pointers

lindenmckenzie:
Demo PR.

### DIFF
--- a/models/users.go
+++ b/models/users.go
@@ -72,25 +72,23 @@ func (p UsersList) BuildListUserRequest(filterString, requiredAttribute string, 
 }
 
 // MapCognitoUsers maps the users from the cognito response into the UsersList Users attribute and sets the Count attribute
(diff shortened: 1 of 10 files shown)


models/users.go
```diff
@@ -112,17 +110,17 @@ type UserParams struct {
...
+		return NewError(ctx, err, InternalError, PasswordGenerationErrorDescription)
 	}
 	p.Password = tempPassword
-	return p.Password, nil
+	return nil
 }
 
 // ValidateRegistration validates the required fields for user creation, returning validation errors for any failures
-func (p UserParams) ValidateRegistration(ctx context.Context, allowedDomains []string) []error {
+func (p *UserParams) ValidateRegistration(ctx context.Context, allowedDomains []string) []error {
```
cookel2, line 123: This line now gives the error `Struct UserParams has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation.` That's the whole problem.

cookel2 (approved):

PR #239, opened 2025-06-26
Feature/v2 linter

lindenmckenzie:
### What

- Upgrade Go
- Upgrade linter to V2
- Upgrade deps

### How to review

Check looks ok, tests pass.

### Who can review

Not me.


redhug1 (commented):
See single comment ...

ci/scripts/lint.sh
```diff
@@ -5,6 +5,6 @@ cwd=$(pwd)
 npm install -g @redocly/cli
 
 pushd $cwd/dp-identity-api
-  go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8
+  go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6
```
redhug1, line 8: would this benefit from the change you made in this PR:

https://github.com/ONSdigital/dp-image-api/pull/96/files

to use container with lint built in ...

lindenmckenzie: sadly not as this one needs node as well and I haven't done a node-go-linter image.

redhug1: Hmm, node as well ... that'll take even longer to pull ?

redhug1 (approved):
LGTM

ci/scripts/lint.sh
lindenmckenzie: so the image has node already - just not the linter.

PR #245, opened 2025-09-09
Remove reset user script

lindenmckenzie:
### What

Removed reset user script in favour of API call.

### How to review

Check I've deleted the right thing.

### Who can review

Not me.


dhee-tree (approved):
Removed only reset password script

PR #246, opened 2025-09-29
Add dis-vulncheck

cookel2:
### What

Jira ticket: https://officefornationalstatistics.atlassian.net/jira/software/c/projects/DIS/boards/1824?selectedIssue=DIS-3854

Upgrading to use dis-vulncheck instead of nancy. This is to avoid the problem of authenticating into OSS Index.

### How to review

Sense check and make sure that tests pass.

### Who can review

Anyone in Open Sauce.


Dockerfile.local
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24.4-bullseye AS base
+FROM golang:1.24.7-bookworm AS base
```
lindenmckenzie:
```suggestion
FROM golang:1.24 AS base
```

By only specifying the minor version, it makes the image more reusable for local docker stacks.

lindenmckenzie (commented):
Suggestion for the Dockerfile.local

Dockerfile.local
cookel2: That makes sense - done 👍

mr-nick17 (approved):

